PR #20, opened 2023-12-07
Use subplot mosaic instead of gridspec

AlecThomson:
Using the `subplot_mosaic` API over `gridspec` for validation plots.

Still need to test this to see if actually looks good.

Might also be nice to add the filed info whilst we're at it.


AlecThomson:
I'm now pretty happy with the plotting state. I've left some TODOs for the extra info plot

Here's an example with some dummy data:
test_val

AlecThomson:
Added some better placeholders

PR #39, opened 2024-01-11
Validation tables

AlecThomson:
Adds the ability to produce the validation tables.

Still some missing data to add to the validation, notably per-beam PSF information.

Will also need to link into artefact creation.


AlecThomson:
Added table artifacts

tjgalvin:
I am going to merge this in now as it is becoming a little too much of a single change. 

The to-dos are:
- correct the place holder psf holders
- update the `MSSummary` class to hold the items that the field panel  expects

The second option should be straight forward, and should allow the measurement sets to be self-cal'd and zipped again without having to wait for the validation plots / tables,

Issue #53, opened 2024-01-22
Stokes V imaging


Sunmish:
Issue to note requirements for Stokes V imaging + validation/leakage assessment/etc.

Stokes V images would use the same imaging settings as Stokes I, but done separately. This requires both Stokes I and V output image names to be unique - add e.g. ".i"/".v" to image names.

Leakage extraction for validation of residual leakage requires Stokes I source list and Stokes V image. 

Example leakage extraction script below, though an alternate approach might be adding Stokes V column to aegean-output source lists.

```python
#!/usr/bin/env python

import numpy as np

from astropy.io import fits
from astropy.table import Table
from astropy.coordinates import SkyCoord
from astropy import units as u
from astropy.wcs import WCS
from astropy.stats import sigma_clip

KEEP_COLUMNS = {
    "ra": "StokesI_RA",
    "dec": "StokesI_DEC",
    "peak_flux": "StokesI_flux_peak",
    "int_flux": "StokesI_flux_int",
    "local_rms": "StokesI_image_rms",
}

class StokesData():
    """Class to extract and hold Stokes Q, U, or V leakage data."""
    def __init__(self, catalogue_i, image_p,
        ra_key="ra",
        dec_key="dec",
        multiplier=1.):
        """
        Parameters
        ----------
        catalogue_i : str
            Filename for Stokes I catalogue. Normally the Selavy catalogue.
        image_p : str
            Filename for Stokes Q, U, or V image.
        ra_key : str, optional
            Name of RA key in catalogue. [Default 'ra']
        dec_key : str, optional
            Name of dec. key in catalogue. [Default 'dec']
        multiplier : float, optional
            Value to multiply catalogue flux densities by. Note for Selavy this should
            be set to 1e-3 as Selavy by default reports in mJy. [Default 1.0]

        """

        if catalogue_i.lower().endswith(".fits"):
            table_format = "fits"
        elif catalogue_i.lower().endswith(".vot") or catalogue_i.lower().endswith(".xml"):
            table_format = "votable"
        else:
            table_format = None

        self.catalogue_i = Table.read(catalogue_i, format=table_format)
        self.image_p = fits.open(image_p)
        self.header = self.image_p[0].header
        self.wcs = WCS(self.header).celestial
        self.data = np.squeeze(self.image_p[0].data)
        self.multiplier = multiplier
        abs_clipped = sigma_clip(np.abs(self.data), 
            sigma=5, 
            maxiters=3, 
            masked=False, 
[...]

PR #54, opened 2024-01-22
add leakage extraction script with associated command line interface

Sunmish:
Second part of issue #53. 

Simple leakage extraction using Stokes I catalogue and polarization image (default Stokes V), also assuming aegean source-finding is used.


tjgalvin:
Addressed (mostly) in #157

PR #73, opened 2024-04-02
Flagging

AlecThomson:
Adds a flagging summary sub-plot to the overall summary plot.

I've also made some small tweaks to catch brittle errors that arise when using test data (i.e. where some values are `None`). We may want to add some extra data to the test directory so that the full summaries can be generated for a test run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,4 +126,5 @@ dmypy.json
 
 #
 flint/data/tests/*.ms
-flint/tests/*.csv
+tests/*.csv
+tests/*.png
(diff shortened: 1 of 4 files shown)


AlecThomson:
Here's an example of the new plot format
validation_39400-RACS_0635-31

tjgalvin:
All of this looks reasonable to me. 

The only major issue I see is that some formatting should go into the legend of the flagging figure. I think it will very quickly cover the field image when all 36 beams are inspected.

tjgalvin:
@AlecThomson  - I have added some small tweaks. Do you have any thoughts? I am just going to yolo it otherwise

AlecThomson:
LGTM. Do you have a demo of the new plot?

tjgalvin:
No, hence the yolo'ing

PR #80, opened 2024-04-11
Added a template approach to specify imaging and self-calibration settings

tjgalvin:
To-date imaging (e.g. wsclean), self-calibration (e.g. gaincal) and other things have been hard-coded into the `flint` codebase. When updating these throughout the rounds of self-calibration a dictionary with updated values would be obtained and splatted into the creation of the appropriate `Options`. 

This is pretty tiresome. If the `git` history is examined it would be pretty obvious how many times commits were made purely to change the image size, solution interval, or something trivial. 

This pull request introduces a template yaml file that is used to set the options that may or may not evolve as the pipeline runs. The idea is that a file (at the moment a yaml file) is parsed into a `Strategy` class, which is nothing more than a glorified name for a dictionary (but nice for type checking). Throughout the pipeline the `Strategy` with be investigated to:
- get defaults for an `Option` mode
- check for and updated values for the specific round of imaging
- update the defaults
- return the updated defaults

The existing infrastructure to overload the hardcoded defaults of an `Option` type class is used as normal.

### DIFF
--- a/flint/configuration.py
+++ b/flint/configuration.py
@@ -341,6 +341,25 @@ def load_strategy_yaml(input_yaml: Path, verify: bool = True) -> Strategy:
     return input_strategy
 
 
+def write_strategy_to_yaml(strategy: Strategy, output_path: Path) -> Path:
+    """Write the contents of a current strategy to a yaml file
+
+    Args:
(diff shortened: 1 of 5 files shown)


tjgalvin:
This addresses #4

Issue #81, opened 2024-04-11
Copying strategy file into output science folder


tjgalvin:
If a strategy configuration file is supplied via the `--imaging-strategy` option when calling the continuum pipeline, this should be copied to the output science directory and then loaded.

PR #83, opened 2024-04-12
Copy strategy file to science directory

tjgalvin:
If a strategy file is used, this pull request will copy a timestamped version into the output directory of the science data.


tjgalvin:
Addresses #81

Issue #82, opened 2024-04-11
Move to configargparse for main pipelines


tjgalvin:
It would be beneficial to use the droping in replace for `argparse` that uses configuration files in place of strictly CLI arguments. This would be useful to also dump into the output science directory.

tjgalvin:
Performed in #84

tjgalvin:
Actually, although the configuration argparse is now in place it should still be dumped to the output file. Will investigate.

tjgalvin:
Upon thinking about it, it should probably actually be the `FieldOptions` structure that is passed into the workflow that is dumped. Closing this.

PR #84, opened 2024-04-13
Add Configargparse to main continuum pipeline

tjgalvin:
It is nice to have a way of providing a configuration file to list the optional CLI arguments for the main continuum imaging and self-calibration pipeline. 

I have added `configargparse` as a dependency in order to allow a user to either:
- specify optional CLI arguments normally
- set a configuration file that outlines all optional arguments
- set a configuration file for optional arguments which may be obverloaded on the CLI

It is a drop in replacement for the normal `argparse.ArgumentParser` and seems to be working nicely. A primitive unit test has also been added.


tjgalvin:
Addresses #82

Issue #64, opened 2024-02-13
Add potato to peel sources


tjgalvin:
@sunmish has developed a tool named `potato` to peel out those annoying sources and has recently containerised it. Some work has previously gone into mocking up something to draw it into flint and the typical continuum pipeline. Now that the container is working this should be revisited. 

There is a separate branch named `peel` (I think) that has the mock up and initial functions to evaluate whether a known source should be peeled.

tjgalvin:
Performed in #92. This seems to be working, and is a matter of configuration / settings

PR #92, opened 2024-04-25
Add potato peel to continuum imaging

tjgalvin:
Add in the initial infrastructure to support peeling reference sources via potato peel.


tjgalvin:
Addresses #64

Issue #86, opened 2024-04-15
Add rotation to MSSummary


tjgalvin:
The ASKAP instrument has a third axis, which is typically fixed to the sky throughout an observation. This information is recorded int he MS, and is used by `fixms` in order to correct the polarimetry for use in wsclean. 

This original rotation should be recorded in a MSSummary object, which could come in handy when estimating the apparent sky brightness via the holography beams (which are not rotationally invariant).

tjgalvin:
I am working through this under the [the rotation branch](/tjgalvin/flint/tree/rotation)

tjgalvin:
And I have merged this branch under #96

PR #96, opened 2024-05-02
Rotate holography beams when linmos'ing

tjgalvin:
The ASKAP third-axis (physical rotation) and rotation of the footprint (electronically rotated) together will place the instrumental response onto the sky in a particular direction. If the science field has a different alignment than the data collected during holography, there needs to be an alignment process when coadding images in order to properly apply the constrained holography beams to the same direction as the science field. 

This pull request aims to:
- pull out the pol axis from a measurement set
- carry this information through the pipeline
- pass it over to linmos, deriving the appropriate rotation using *an assumed* parameterisation of the holography beam alignment

This aims to address #86

### DIFF
--- a/flint/coadd/linmos.py
+++ b/flint/coadd/linmos.py
@@ -13,7 +13,7 @@ from flint.sclient import run_singularity_command
 
 # This is the expected orientation of the third-axis and footprint (remember the footprint
 # can be electronically rotated).
-EXPECTED_HOLOGRAPHY_ROTATION_CONSTANT_RADIANS = -np.pi / 8
+EXPECTED_HOLOGRAPHY_ROTATION_CONSTANT_RADIANS = -2 * np.pi / 8
 
 
 class LinmosCommand(NamedTuple):
--- a/flint/coadd/linmos.py
+++ b/flint/coadd/linmos.py
@@ -248,8 +248,8 @@ def generate_linmos_parameter_set(
     logger.info(f"{len(img_str)} unique images from {len(images)} input collection. ")
     img_list: str = "[" + ",".join(img_str) + "]"
 
-    assert (
-        len(set(img_str)) == len(images)
+    assert len(set(img_str)) == len(
+        images
     ), "Some images were dropped from the linmos image string. Something is bad, walk the plank. "
 
     # If no weights_list has been provided (and therefore no optimal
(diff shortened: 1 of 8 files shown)


flint/summary.py
```diff
@@ -89,6 +92,26 @@ def with_options(self, **kwargs) -> FieldSummary:
         return FieldSummary(**prop)
 
 
+def _get_pol_axis_as_rad(ms: Union[MS, Path]) -> float:
```
tjgalvin, line 95: This is the function that attempts to get the most correct orientation. I am not sure whether I want to force an error to be raised if the instrument receptor angle is missing, or whether this is ok. This is a non-standard keyword.

flint/summary.py
```diff
@@ -126,11 +149,16 @@ def add_ms_summaries(
     hour_angles = centre_altaz.az.to(u.hourangle)
     elevations = centre_altaz.alt.to(u.deg)
 
+    # The INSTRUMENT_RECEPTOR_ANGLE comes from fixms and is
```
tjgalvin, line 152: Once it has been added to the field summary, then it is passed through to the linmos tasks.

flint/prefect/flows/continuum_pipeline.py
```diff
@@ -251,98 +253,104 @@ def process_science_fields(
     fits_beam_masks = None
 
     for current_round in range(1, field_options.rounds + 1):
```
tjgalvin, line 255: Much of this has changed simply from the `with tags` context manager. But do note passing in the field summary to the appropriate linmos tasks

flint/coadd/linmos.py
```diff
@@ -259,6 +267,19 @@ def generate_linmos_parameter_set(
     # attribute drops the parent directory (absolute directory).
     parent_dir = linmos_names.image_fits.parent
 
+    # Work out the appropriate rotation to provide linmos. This should
```
tjgalvin, line 270: This is the part that most worries me. I think it is correct. But, who knows. Maybe you. You pirate.

flint/summary.py
AlecThomson: Depends on how much we want to be locked into FixMS. Will the observatory MSs ever have correct feeds?

AlecThomson (approved):
Looking all good to me. We may want to consider dumping the field summary to FITS headers at some point...

flint/summary.py
tjgalvin: At this point I would not expect them to.

If they did I would they would also rotate the correlations to the sky as well as update the receptor angle. if so, I think the problem kind of drops out even if we keep fixms in out respective codez.

flint/coadd/linmos.py
```diff
@@ -259,6 +267,19 @@ def generate_linmos_parameter_set(
...
+    # be the differential pol. axis. rotation between the science field
+    # and the holography. At the moment this holography rotation is
+    # unknown to us (not stored in the cube header).
+    alpha = None
+    if pol_axis:
+        logger.info(
+            f"The constant assumed holography pol_axis + footprint rotation is: {EXPECTED_HOLOGRAPHY_ROTATION_CONSTANT_RADIANS:.4f} radians"
```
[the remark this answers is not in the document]
tjgalvin: Perhaps it is a naming thing. I have no tie to the log message, other than just putting something there for reporting. Thoughts on the alternative?

PR #164, opened 2024-08-19
Added aoflagger / flind finalcutoff

tjgalvin:
A couple of unrelated but small changes:
- run aoflagger on casda downloaded measurement sets
- use a new linmos with a `finalcutoff` option


tjgalvin:
Issue with `griffe` removing a dataclasses definition. To be looked at after merge

PR #167, opened 2024-08-27
An idea - Drop casacore

AlecThomson:
Python 3.8 is reaching EOL.

As confirmed in https://github.com/casacore/python-casacore/issues/251, `python-casacore` and modular casa6 (`casatasks` and `casatools`) are not compatible. As also suggested the linked issue, one option could be to just use the underlying casacore functionality from `casatools`. 

The other option would be to just fully containerise casa. I'm just opening this PR to see what the tests think and bounce some ideas.


AlecThomson:
Grr, ok it looks like the API is quite different.

https://casadocs.readthedocs.io/en/v6.5.2/api/tt/casatools.table.html#casatools.table

Table operations require an object to be instantiated, and don't prove a context manager e.g.
```python
from casatools.table import table
tb = table()
tb.open("some_file.ms")
# Do some stuff
...
tb.close()
```
Not a huge fan. Maybe there's something slick to be done with a custom context manager. Needs some extra though

tjgalvin:
I have been switching camps on this - whether to try to use a container of casa and keep python-casacore, or keep the casatasks and figure out how to get to the python-casacore. 

I think long term though the containerise casa and keep python-casacore is probably the safer bet. There are a couple of reasons:
- `casatasks` does not seem to be updated regularly to newer versions of python, while python-casacore is
- other applications `flint` uses have already been set up to use containers, so there its not like it is new territory
- while using the python functions are nicee, ultimately we are not using anything returned from them. That is to say they are just a dumb bucket we through arguments into, and the returns  (if any) are not handled
- I think I'd rather have the python-casacore in all its glory around for other functions/utilities to build from. While I am only using the `table` iterface at the moment, there are definitely other components that I think could be useful (`taql` for instance)

Thoughts?

tjgalvin:
I have gone down the casa in a container route, and largely merged this and other fixes together in #170 . 

Will close this request for the moment. Thanks bigly for helping think through the problem~

PR #205, opened 2025-01-08
Update ruff rules

AlecThomson:
Adds and enables the following `ruff` rulesets:
- [isort](https://docs.astral.sh/ruff/rules/#isort-i)
- [pyupgrade](https://docs.astral.sh/ruff/rules/#pyupgrade-up)
- [NumPy-specific rules](https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy)
- [pandas-vet](https://docs.astral.sh/ruff/rules/#pandas-vet-pd)

I've let stubs if there we want to enable more rules. The biggest one is the `UP` rule, which enables the new-style types.


AlecThomson:
I've also added better spellchecking via `typos` and fixed broken tests

PR #216, opened 2025-01-09
Link to docs

AlecThomson:
Update to point to new docs


AlecThomson:
Also fix up docs that prettier broke

PR #217, opened 2025-01-10
Polarisation cubes

AlecThomson:
Adding a polarisation workflow


tjgalvin:
I think I have all the kinks in the continuum and self-calibration pipeline figured out. The only outstanding question I had is around the difference between `task_linmos_images` and `task_linmos_images_list`.

tjgalvin:
Ok, as far as I see the continuum pipeline is working after a few small tweaks that slipped past ruff. 

I think I could go about accepting this. I'll run one last check and then do that!

Thanks for the massive amount of work @AlecThomson  adding this!

PR #234, opened 2025-02-20
Fixes for pol. imaging

AlecThomson:
A number of fixes based on a real-life test of the polarisation pipeline. It seems like our fearless captain may have accidentally broken things.

Summary of the changes:
- Drop support for `force-mask-rounds`
- Add support for `no-mf-weighting`
- Fix call to `task_linmos_images` in polarisation pipeline

This does really point to a desperate need for running tests on the flow scripts, which we currently do not do. We should discuss seperately.


AlecThomson:
Converted to draft whilst catching some more bugs

AlecThomson:
As discussed in person captain @tjgalvin may want to restore `force-mask` support...

PR #239, opened 2025-02-23
Zeors to NaN in linmos output

tjgalvin:
It has been noted that depending on conditions with the `cutoff` and `finalcutoff` parameters in linmos that there are pixels that are written to the output fits images filled with 0's. These are, presumably, the default values o the internal array. 

This PR introduces a 0.0 to nan step which is used when trimming the linmos images and weights.


tjgalvin:
I forced this one as vizier is currently down.

PR #240, opened 2025-03-04
Spectral cube refactoring

AlecThomson:


tjgalvin:
Thanks @AlecThomson  - I don't have any issues with this. Big thanks as well for moving the prefect tagged tasks into the `flint.prefect` submodule, and getting through the pain of circular imports!

PR #250, opened 2025-03-19
Exposed BANE and aegean options

tjgalvin:
For some time the number of cores and stripes to use in BANE were hardcoded to 8 and 7. These were set a long time ago without much thought. 

Tweaking these numbers would be ideal, especially if larger compute resources are requested per dask worker. 

This change brings in the ability to set the cores through `BANEOptions`. Potentially this could further be expanded to allow it to be auto-detected (may be the default behavior in BANE) but I am a little uncomfortable with this:
1 - there is a bug in current BANE where deadlocking can occur should CPUs == stripes
2 - not sure how the python get cpu functions behave in all environments

I am therefore for the moment in favor of an explicit is better approach. In the continuum pipeline I have also added plumbing to draw from the strategy file. 

Some extensions have been made to `flint_aegean`. 

Added some basic source finder docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@
   - Convert flux columns to Jy
 - Improved docs
 - Added chunking to the NaN/UVW/zero flagging during selfcal
+- Added `update_bane_options` and `update_aegean_options` to `flint.source_finding.aegean`
+  - Initially to support using more cores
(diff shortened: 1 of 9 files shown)


tjgalvin:
I see that the code coverage test is failing, but for the moment without having access to singularity in the test framework I can not really get much more coverage. 

I will force the change.

PR #257, opened 2025-03-24
Docker builds

AlecThomson:
Adds the following:

- A primary Dockerfile for building `flint` containers
- A dev Dockerfile for use with [devcontainers](https://code.visualstudio.com/docs/devcontainers/containers)
- `devcontainer` configuration
- Stubs for testing singularity flows

Still TODO:

- [x] Add GitHub action to build container and release it
- [x] Ensure current testing CI has singularity

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,29 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/docker-existing-dockerfile
+{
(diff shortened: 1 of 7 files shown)


AlecThomson:
@tjgalvin - once we've confirmed that the Docker image build is working as intended we should probably update the install docs with usage instructions. And maybe a warning that a nested container may not be fully stable...

AlecThomson:
I can also add some brief docs on that point to how to use the devcontainer

tjgalvin:
Perhaps we put the new docs in with this PR?

tjgalvin (approved):

PR #262, opened 2025-03-24
Update containers

### LINKED ISSUE #261: Point to new containers package
Need to update `flint_containers` to point to GHCR packages

AlecThomson:
- Updates container URIs to GHCR
- Uses asyncio for downloads
- Adds optional tag argument (defaults to v1.0)

Closes #261

### DIFF
--- a/flint/containers.py
+++ b/flint/containers.py
@@ -3,6 +3,7 @@ containers required for flint"""
 
 from __future__ import annotations
 
+import asyncio
 from argparse import ArgumentParser
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


tjgalvin:
Looks good to me. Approving

Up until now I have not used asyncio as much as I should have. If it introduces a nice speed up without too much confusion on the logger end I am all for it. Only other thought it how this may interact should we ever put this code path in a `@task` scope. Should there be conflicts we might consider dropping it, but for the moment it looks like a good addition! Thanks!

tjgalvin (approved):

Issue #265, opened 2025-03-24
Python 3.13 breaks


AlecThomson:
As found by @Sunmish, Python 3.13 appears to break. We should consider pinning in `pyproject.toml`.

AlecThomson:
Also noting, I'm considering adding a new battery of CI/CD checks that will hopefully test this

PR #266, opened 2025-03-24
Use ci/cd

AlecThomson:
Switch to new CI/CD


AlecThomson:
Cool - as expected Python 3.9, 3.10, and 3.13 fail (as reported in #265). I'll update `pyproject.toml` for now and update the CI

PR #280, opened 2025-03-27
Dockerfile test

AlecThomson:
Add tests for dockerfile


AlecThomson:
Tests failing, as expected. Putting us into red state for now. Fix incoming from @tjgalvin

PR #294, opened 2025-05-01
Add option to rotate cube inplace. Fixes #293

### LINKED ISSUE #293: Some filesystems dont like rotating fits cubes in place during wsclean step
When running flint on Niagara (https://docs.alliancecan.ca/wiki/Niagara) both the /scratch and /project disks are apparently a `Parallel high-performance filesystem (IBM Spectrum Scale)`, but throw the following astropy error when trying to run `flint.imager.wsclean.rotate_cube `

```
  File "/project/b/bmg/eosinga/doradogroup/software/miniforge3/envs/flint/lib/python3.12/site-packages/astropy/io/fits/hdu/base.py", line 716, in _writeto
    with _free_space_check(self, dirname):
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/project/b/bmg/eosinga/doradogroup/software/miniforge3/envs/flint/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/project/b/bmg/eosinga/doradogroup/software/miniforge3/envs/flint/lib/python3.12/site-packages/astropy/io/fits/util.py", line 806, in _free_space_check
    raise OSError(error_message + str(exc))
OSError: 675942912 requested and 0 written
```

A fix would be to expose an option for the selfcal config to allow the user to ask for the rotation to be done not in place but via a temporary file

ErikOsinga:
This exposes the option for the selfcal config file to add `--flint_make_cube_inplace False` to do the fits axis rotation via a temporary file to fix #293 

Possible TODO: it would be nice as well if `rotate_cube` actually checked whether the 3rd and 4th axis are indeed (pol,chan) before swapping them to (chan,pol) so that `rotate_cube` could be made redundant upon re-running.

* Forgive me if I've botched any regular way you all do options/expressions/typing

### DIFF
--- a/flint/imager/wsclean.py
+++ b/flint/imager/wsclean.py
@@ -17,6 +17,7 @@ with a ``.``.
 from __future__ import annotations
 
 import re
+import os
 from argparse import ArgumentParser
 from glob import glob
 from numbers import Number
@@ -179,7 +180,8 @@ class WSCleanOptions(BaseOptions):
     """If True do not log the wsclean output"""
     no_mf_weighting: bool = False
     """Opposite of -ms-weighting; can be used to turn off MF weighting in -join-channels mode"""
-
+    flint_make_cube_inplace: bool = True
+    """Rotate the cube for the linmos axis ordering in place, or do it via a temporary file that then gets deleted. Good thing to turn off when getting weird OSErrors on file writing"""
 
 class WSCleanResult(BaseOptions):
     """Simple container for a wsclean command."""
@@ -211,6 +213,7 @@ def combine_images_to_cube(
     prefix: str,
     mode: str,
     remove_original_images: bool = False,
+    inplace: bool = True,
 ) -> Path:
     """Combine wsclean subband channel images into a cube. Each collection attribute
     of the input `image_set` will be inspected. The MFS images will be ignored.
@@ -220,7 +223,8 @@ def combine_images_to_cube(
     Args:
         image_set (ImageSet): Collection of wsclean image productds
         remove_original_images (bool, optional): If True, images that went into the cube are removed. Defaults to False.
-
+        inplace (bool, optional): If True, modify the file in-place. If False, write to a temporary file and
+        then replace the original. Default True
     Returns:
         ImageSet: Updated iamgeset describing the new outputs
     """
@@ -230,7 +234,7 @@ def combine_images_to_cube(
 
     logger.info(f"Combining {len(images)} images. {images=}")
     freqs = combine_fits(file_list=images, out_cube=output_cube_name)
-    rotate_cube(output_cube_name)
+    rotate_cube(output_cube_name, inplace=inplace)
 
     # Write out the hdu to preserve the beam table constructed in fitscube
     logger.info(f"Writing {output_cube_name=}")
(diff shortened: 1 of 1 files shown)


flint/imager/wsclean.py
```diff
@@ -882,34 +887,67 @@ def create_wsclean_cmd(
...
-            new_header[f"CRVAL{a}"] = tmp_header[f"CRVAL{b}"]
-            new_header[f"CDELT{a}"] = tmp_header[f"CDELT{b}"]
-            new_header[f"CUNIT{a}"] = tmp_header[f"CUNIT{b}"]
-
-        # Cube is currently STOKES, FREQ, RA, DEC - needs to be FREQ, STOKES, RA, DEC
-        data_cube = np.moveaxis(data_cube, 1, 0)
-        hdu.data = data_cube
-        hdu.header = new_header
-        hdulist.flush()
+def rotate_cube(output_cube_name: str, inplace: bool = True) -> None:
```
AlecThomson, line 890: Nice catch on the type hint! Rather than just a string, let's use a `str | Path` and cast immediately to `Path`. Then we can use `Path` methods over the `os` module

flint/imager/wsclean.py
```diff
@@ -882,34 +887,67 @@ def create_wsclean_cmd(
...
+    ----------
+    output_cube_name : str
+        Path to the FITS cube to rotate.
+    inplace : bool, optional
+        If True, modify the file in-place. If False, write to a temporary file and
+        then replace the original. Default True
+    """
+    # Determine temporary path if not inplace
+    if not inplace:
+        base, ext = os.path.splitext(output_cube_name)
```
AlecThomson, line 908: Use `.name` and `.suffix` instead

flint/imager/wsclean.py
```diff
@@ -882,34 +887,67 @@ def create_wsclean_cmd(
...
+            hdul[0].data = rotated_data
+            hdul[0].header = header
+            hdul.flush()
+    else:
+        # Write to temporary file
+        logger.info(f"Writing rotated cube to temporary file {tmp_name}")
+        hdu = fits.PrimaryHDU(data=rotated_data, header=header)
+        hdu.writeto(tmp_name, overwrite=True)
+        # Replace original file
+        os.remove(output_cube_name)
```
AlecThomson, line 942: Use `.unlink`

flint/imager/wsclean.py
```diff
@@ -882,34 +887,67 @@ def create_wsclean_cmd(
...
+        os.rename(tmp_name, output_cube_name)
```
AlecThomson, line 943: `.rename`

flint/imager/wsclean.py
AlecThomson, line 890: I also forgot our own rule of returning a value. We should probably return the `Path` object on success. That also means we can return early on given logical branches

AlecThomson (changes requested):

AlecThomson:
Thanks for the PR @ErikOsinga! Don't forget to install the dev tools (like pre-commit) to help with the auto-formatting etc.

ErikOsinga:
Ehh, hope that worked!

AlecThomson (approved):

Issue #295, opened 2025-05-02
Can't set wsclean temp_dir with environment variable as part of a larger path.


ErikOsinga:
Currently the function `create_wsclean_cmd` crashes with

```
AssertionError: name_dir=None is None, which is bad
```

when a wsclean temp_dir is set in the selfcal configuration file that's a path which uses an environment variable, rather than just an environment variable, e.g. when it's `"$SLURM_TMPDIR/wsclean_tmpdir/"` instead of `$SLURM_TMPDIR`

![Image](https://github.com/user-attachments/assets/2b5e0b6d-739d-4d86-87bf-ac37b4113837)

It'd be nice to enhance the functionality of `flint.utils.get_environment_variable` to handle such cases

AlecThomson:
Closing for now. Let us know if this still breaks for you @ErikOsinga and we can reopen

ErikOsinga:
Seems to be working, appreciate the quick turnaround :)

PR #296, opened 2025-05-02
Proposed more flexible handling of environment variables if the user wants to mix those with paths

ErikOsinga:
Addresses #295, now `get_environment_variable()` attempts to resolve each part of a path as a possible environment variable, instead of only assuming that the full path is an environment variable.

Example behaviour (updated test):

```
@pytest.fixture(scope="session", autouse=True)
def set_env():
    """Set up variables for a specific test"""
    os.environ["TEST1"] = "Pirates"
    os.environ["TEST2"] = "Treasure"


def test_get_environment_variable(set_env):
    """Make sure that the variable is processed nicely when getting environment variable"""
    val = get_environment_variable("TEST1")
    assert val == "Pirates"
    val2 = get_environment_variable("$TEST2")
    assert val2 == "Treasure"
    val3 = get_environment_variable("THISNOEXISTS")
    assert (
        val3 == "THISNOEXISTS"
    )  # in case the user gives a path that isnt a variable return the same thing
    val4 = get_environment_variable("TEST1/TEST2")
    assert val4 == "Pirates/Treasure"
    val5 = get_environment_variable("TEST1/TEST2/TEST3")
    assert (
        val5 == "Pirates/Treasure/TEST3"
    )  # in case the user gives a path that includes variables and directories
```

### DIFF
--- a/flint/imager/wsclean.py
+++ b/flint/imager/wsclean.py
@@ -17,6 +17,7 @@ with a ``.``.
 from __future__ import annotations
 
 import re
+import os
 from argparse import ArgumentParser
 from glob import glob
 from numbers import Number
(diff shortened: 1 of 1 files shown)


ErikOsinga:
something went wrong here with the branches, ignore this one....

PR #297, opened 2025-05-02
Proposed more flexible handling of environment variables if the user wants to mix those with paths

ErikOsinga:
Addresses #295, now `get_environment_variable()` attempts to resolve each part of a path as a possible environment variable, instead of only assuming that the full path is an environment variable.

Example behaviour (updated test):

```
@pytest.fixture(scope="session", autouse=True)
def set_env():
    """Set up variables for a specific test"""
    os.environ["TEST1"] = "Pirates"
    os.environ["TEST2"] = "Treasure"


def test_get_environment_variable(set_env):
    """Make sure that the variable is processed nicely when getting environment variable"""
    val = get_environment_variable("TEST1")
    assert val == "Pirates"
    val2 = get_environment_variable("$TEST2")
    assert val2 == "Treasure"
    val3 = get_environment_variable("THISNOEXISTS")
    assert (
        val3 == "THISNOEXISTS"
    )  # in case the user gives a path that isnt a variable return the same thing
    val4 = get_environment_variable("TEST1/TEST2")
    assert val4 == "Pirates/Treasure"
    val5 = get_environment_variable("TEST1/TEST2/TEST3")
    assert (
        val5 == "Pirates/Treasure/TEST3"
    )  # in case the user gives a path that includes variables and directories
```

### DIFF
--- a/flint/imager/wsclean.py
+++ b/flint/imager/wsclean.py
@@ -179,6 +179,8 @@ class WSCleanOptions(BaseOptions):
     """If True do not log the wsclean output"""
     no_mf_weighting: bool = False
     """Opposite of -ms-weighting; can be used to turn off MF weighting in -join-channels mode"""
+    flint_make_cube_inplace: bool = True
+    """Rotate the cube for the linmos axis ordering in place, or do it via a temporary file that then gets deleted. Good thing to turn off when getting weird OSErrors on file writing"""
 
 
 class WSCleanResult(BaseOptions):
@@ -211,6 +213,7 @@ def combine_images_to_cube(
     prefix: str,
     mode: str,
     remove_original_images: bool = False,
+    inplace: bool = True,
 ) -> Path:
     """Combine wsclean subband channel images into a cube. Each collection attribute
     of the input `image_set` will be inspected. The MFS images will be ignored.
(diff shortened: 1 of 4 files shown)


ErikOsinga:
It should now leave integer arguments alone as before, e.g. `size 1024` is correctly parsed again by `_resolve_wsclean_key_value_to_cli_str`

tjgalvin:
I have added another test for the trailing slash. 

Looking at this now I am unsure why I am allowing environment variables without a '$' to be resolved. I am thinking that `os.getenv` should only be called if `part` starts with `$`.

AlecThomson:
Also noting that codecov is reporting that returning `None` is not tested, nor is returning `default`

tjgalvin:
I have added some extra tests and re-arranged the `get_envirnment_variable` function. 

This is about as far as I think we want to go in terms of flexibility for this lookup. I see this becoming a bit of a foot-gun. We are starting to treat / assume that this is a path type value being resolved, and building that assumption that goes outside a simple environment variable lookup.\ For the moment I think we have found a nice balance, but much more should go out to a separate function.

tjgalvin:
@AlecThomson  - if you are happy with these feel free to approve + merge

AlecThomson:
Hey all, this is looking all gucci to me. My only thought is the name of the function. This feel more like a:
```
parse_environment_variables(...)
```
Since it's got a straight `get`, and it can parse multiple variables. But not a huge deal ultimately

AlecThomson (approved):

Issue #303, opened 2025-05-06
KeyError: 'default' in hot_potato


ErikOsinga:
In the `flint_flow_continuum_pipeline` the task `hot_potato` doesn't find the configuration file, resulting in KeyError `default` even though the configuration file exists on disk.

![Image](https://github.com/user-attachments/assets/4eaa0773-e13d-4c43-8db6-b7cee78b333c)

This is because the directory is not bound correctly when the singularity command is constructed

![Image](https://github.com/user-attachments/assets/71795e2d-08c4-48e0-8f89-c366ecc5cc5d)

Notice how it does

```
  --bind /project/b/bmg/eosinga/doradogroup/skymodel_prelim_with_BB/peel \
  --bind /bb/b/bmg/eosinga/selfcaloutput/69817/SB69817.NGC1566_band1.beam02.ms \
``` 
but then tries to 
```
  -c /bb/b/bmg/eosinga/selfcaloutput/69817/SB69817.NGC1566_band1.beam02.potato.config \
```

resulting in the bad descripted "KeyError" , while in fact it should be a FileNotFoundError. This is due to the slightly annoying behaviour of `config.read` not checking whether a file exists, but always succeeding. Could be caught in [`hot_potato](https://gitlab.com/Sunmish/potato/-/blob/main/scripts/hot_potato) to raise a more descriptive error (issue raised there). 

![Image](https://github.com/user-attachments/assets/1d539834-897f-42df-9740-85924d3e1455)

A fix seems to be to update `create_run_potato_peel` to have the parent of the measurement set included in the bind dirs by default, because the config file is always written w.r.t the measurement set location. 

```
    bind_dirs = [
        ms.path,
        ms.path.parent
    ] 
```

This also raises the more complicated question whether it'd be nice to expose some `potato_peel` options to the user, like where to save the config file, or whether to set a custom `.T` directory (Where the temporary wsclean files will be written to), because now the default is the directory `./peel` which is not the best location if a user is trying to run things on a different/faster disk.

AlecThomson:
Closed by #304

PR #304, opened 2025-05-06
Fix issue of potato config file not being found due to missing binding

ErikOsinga:
Addresses [#303](https://github.com/flint-crew/flint/issues/303)


ErikOsinga:
Added a test for `create_run_potato_peel` as well, though this doesn't test whether the actual singularity container works, I think that's covered somewhere else?

AlecThomson (approved):
Adding tests is the pirate way, yarr. Ye be a legendary sailor

PR #305, opened 2025-05-06
Add option --potato-peel-options to continuum flow

ErikOsinga:
Add option `flint_flint_continuum_pipeline --potato-peel-options [yamlfile]` for user to input custom instructions to hot_potato. Allows e.g. changing of tmpdir for potato peel. 

An example `peel_options.yaml` file is uploaded to `flint/data/peel/peel_options.yaml` and also used in testing of this functionality.

1. Do we want to have this file in `data/tests/peel_options.yaml` as well? 

2. Do we potentially want more functionality with this yaml file? We could think of the file consisting of different headers like e.g. "potato_peel_options", "potato_config_options"

### DIFF
--- a/flint/data/peel/peel_options.yaml
+++ b/flint/data/peel/peel_options.yaml
@@ -1,0 +1,10 @@
+# potato_peel_defaults.yaml
+
+# Whether to use direct model subtraction (no self-cal) for faint sources
+direct_subtract: true
+
+# Create an image after each calibration/subtraction loop
+intermediate_peels: true
+
+# Directory for temporary wsclean files
+T: "peel"
--- a/flint/peel/potato.py
+++ b/flint/peel/potato.py
@@ -23,6 +23,7 @@ from typing import Any, Collection, NamedTuple
 
 import astropy.units as u
 import numpy as np
+import yaml
 from astropy.coordinates import SkyCoord
 from astropy.table import Table
 from casacore.tables import table
@@ -38,6 +39,16 @@ from flint.utils import (
     generate_strict_stub_wcs_header,
     get_packaged_resource_path,
 )
+
+
+# A simple representation to pass around. Will help the type
+# analysis and future pirates be clear with their mutinous
+# intentions
+class PotatoPeelStrategy(dict):
+    """Base representation for handling a loaded PotatoPeelStrategy
+    strategy"""
+
+    pass
 
 
 class PotatoConfigOptions(NamedTuple):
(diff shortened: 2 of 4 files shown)


AlecThomson:
This is looking very nice @ErikOsinga! @tjgalvin may want to look at the options construction, as he has stronger thoughts on that kind of architecture than I do. It looks like you've borrowed across the constructor from the other options classes, right?

@tjgalvin - we may want to look at providing an `AbstractBaseClass` for this kind of thing if we're going to be re-using this kind of functionality across the project.

Whilst we're here, are we happy to make a change to the potato option? We're passing an option right now as `T` (which we inherit from potato) but its quite a confusing name. Further, it looks like the canonical name for a `TypeVar`. I see that potato has the other name for this as `--tmp`, could we use that instead?

@ErikOsinga - merging #304 has caused a merge conflict with this branch. Sorry for that.

tjgalvin:
Hi all - thanks for putting this together! 

A couple of thoughts and suggestions, it might be better as a separate pull request. 

- I agree with the renaming of `T` to the long form of the argument, `--tmp`. 
- 
- I do think exposing the PotatoPeelOptions to the user. However, I don't think the approach in this PR is the ideal way. What I suggest is adding the `PotatoPeelOptions` and `PotatoPeelConfigOptions` into the strategy interface. It should be as simple as adding the classes to the recognized set of other Options classes. Then we can use the `get_options_from_strategy` function to pull them out of the strategy yaml file. 

- I'd prefer adding new argparse options to the `FieldOptions` class instead of adding them to the `ArgumentParser`. Doing it this way would mean that the extra option makes it way into the flow scope without having to add extra keyword arguments to the setup flow function and the flow function itself. 

Does that make sense?

Binding to the parent directory of the MS is something that should also be done, without question. A more meaningful tmp directory is also something I'd be more than happy to accept. I set this up a long time ago and have not looked back. Perhaps the thing to do is to set `PotatoPeelOption.tmp = ms.path.parent / "peel"` should `tmp` b None. 

Thoughts?

ErikOsinga:
Thanks for your thoughts on it. I like the idea of having it as part of the selfcal strategy file, to also minimize the amount of different config files a user has to make. 

Just to make sure I understand it correctly, you are saying that flint should be able to resolve a selfcal strategy file that looks e.g. like this:

```
version: 0.2
defaults:
  wsclean:
    abs_mem: 100
    local_rms_window: 65
    size: 10128
selfcal:
  0:
    wsclean: {}
  1:
    wsclean:
      data_column: "EXAMPLE"
    gaincal: {}
peel:
  direct_subtract: true
  intermediate_peels: true
  # directory for tmp files
  tmp: "peel"
``` 

and that this is handled by the `FieldOptions` class?


I also prefer the suggestion for `PotatoPeelOption.tmp = ms.path.parent / "peel"` should if `tmp is None`.

If so, I can create a new PR to implement that.

tjgalvin:
I have mocked this up in the `potato_peel` branch, but it is not tested. 

I have changed `T` to `tmp`. 

@ErikOsinga - yes, that is basically right. It will end up being this in the strategy file:

```yaml
version: 0.2
defaults:
  wsclean:
    abs_mem: 100
    local_rms_window: 65
    size: 10128
selfcal:
  0:
    wsclean: {}
    potatopeel:
      direct_subtract: true
      intermediate_peels: true
      # directory for tmp files
      tmp: null
  1:
    wsclean:
      data_column: "EXAMPLE"
    gaincal: {}
```


I set `tmp` to `null` so that the routine that drives the potato peeling will insert its own `tmp` path, with the behavior being what you suggested:

```python
tmp_dir = Path(ms.path.parent) / "peel"
potato_peel_options = potato_peel_options.with_options(tmp=tmp_dir)
```

I have not tested this. Perhaps if you want to you could also write a simple unit test to make sure the getter is getting from an example strategy file as we expect? Not a huge deal as it is going through all the same existing code that is being tested anyway.

tjgalvin:
I have opened #308  for further discussion.

PR #306, opened 2025-05-06
fix beam summaries racing against gaincal

ErikOsinga:
I think this should fix the issue of the task `create_beam_summary` failing because `gaincal_applycal` has already renamed the MS


AlecThomson (approved):
Nice catch!

tjgalvin:
Nice catch. Yes, there is a little race condition here that I have not generally hit. It must be because of imaging the single beam. This whole flow should really be redesigned to help track such potential events. Looking over the code now.

PR #315, opened 2025-05-21
Rename mbc to mac

tjgalvin:
The masking procedure has been renamed `mbc` (minimum box clip) to the much cooler `mac` (minimum absolute clip). 

I thought this was done some time ago, but I am guessing it was in a branch that was never merged. 

In any case. This may be a slightly annoying change as strings in the strategy files need to be updated. But I am OK with this. Sorry to others who are not. This is not a change to the strategy file itself, just to fields in the `MaskingOptions` class.


AlecThomson:
Yes, Mac

Issue #308, opened 2025-05-08
Potato Peel Options in strategy file


tjgalvin:
Following the discussion from #305 

I have mocked this up in the `potato_peel` branch, but it is not tested. 

I have changed `T` to `tmp`. 

@ErikOsinga - yes, that is basically right. It will end up being this in the strategy file:

```yaml
version: 0.2
defaults:
  wsclean:
    abs_mem: 100
    local_rms_window: 65
    size: 10128
selfcal:
  0:
    wsclean: {}
    potatopeel:
      direct_subtract: true
      intermediate_peels: true
      # directory for tmp files
      tmp: null
  1:
    wsclean:
      data_column: "EXAMPLE"
    gaincal: {}
```

I set `tmp` to `null` so that the routine that drives the potato peeling will insert its own `tmp` path, with the behavior being what you suggested:

```python
tmp_dir = Path(ms.path.parent) / "peel"
potato_peel_options = potato_peel_options.with_options(tmp=tmp_dir)
```

I have not tested this. Perhaps if you want to you could also write a simple unit test to make sure the getter is getting from an example strategy file as we expect? Not a huge deal as it is going through all the same existing code that is being tested anyway.

AlecThomson:
If we're adding a new set of options, we should bump the config version, methinks

tjgalvin:
I think that this is more a super set of the existing strategy. That is to say that the behaviour has not changed. All existing strategy files should be oblivious to this. With that in mind I am not sure. I'd definitely agree if there was something that was backwards breaking

tjgalvin:
I had a little time to think through this a little more since yesterday. 

We ought to definitely bump the package version with a new release should this get merged. If a strategy files were to use the `potatopeel` options but the flint package was not updated to a version with this registered `PotatoPeelOptions` the file will fail to verify. 

Updating the `version` string in the strategy file is also an option. I suppose my hesitation is around what the meaning would be. The 0.1 -> 0.2 transition was there was. backwards breaking change - that whole 'initial' keyword section was dropped. In this case I don't think we have such a backwards breaking change as nothing was done to the schema. 

Open to both, but would prefer the package version update, simply as all my current strategy files would work without change.

tjgalvin:
Closed with #316 and new version release

PR #316, opened 2025-05-21
Add PotatoPeelOptions to known strategy options

tjgalvin:
It was requested to expose a way of adding `potato` related options to the user. This PR adds `potato` related options to the set of known `Options` classes used by the strategy file.


tjgalvin:
@ErikOsinga  - would you mind testing this? Should it do what you need to we can merge this, close the other PR, and close out #308

ErikOsinga:
Yup, i'll have a look at this today

flint/peel/potato.py
```diff
@@ -640,15 +647,27 @@ def potato_peel(
...
-    update_potato_config_options = (
-        update_potato_config_options if update_potato_config_options else {}
-    )
-    potato_config_options = PotatoConfigOptions(**update_potato_config_options)
-    potato_config_command = create_run_potato_config(
-        potato_container=potato_container,
-        ms_path=ms,
-        potato_config_options=potato_config_options,
-    )
+    # Should the user specify the potato peel configuration to use we
```
AlecThomson, line 650: Would be good to test this, methinks

AlecThomson: Codecov seems to agree

tjgalvin: Let me looksie

tjgalvin: Ok, I see it a little more clearly now. 

I can pull this bit out and put these steps into an isolated function that makes it a little more straight forward to test. The trouble is that it calls out to singularity to run the potato config creation step.

tests/test_potato.py
```diff
@@ -335,3 +336,44 @@ def test_create_run_potato_peel_with_custom_tmp(tmp_path, ms_example, monkeypatc
     # verify singularity invocation was with our container and the generated command
     assert run_call["image"] == potato_container
     assert run_call["command"] == cmd.command
+
+
+def test_prepare_potato_options(tmp_path, ms_example, monkeypatch):
+    """See if we can generate the correct set of options"""
+    # set up a fake container and MS
+    potato_container = tmp_path / "potato.sif"
+    potato_container.write_text("")  # doesn't actually get used
```
ErikOsinga, line 345: why is this line here?

ErikOsinga:
Actually did find a few bugs,

1. It was calling hot_potato with `-tmp` , but hot_potato only takes `-T` or `--tmp`. 
2. If the user didn't set the `tmp` directory it was failing because the in-place value assignment was missing `potato_peel_options = potato_peel_options.with_options(tmp=tmp_dir)` 
3. `create_run_potato_peel` was creating the command at the start of the function, so the actual changing of the options wasn't propagated to `potato_peel_command`. 

Added a test in `test_create_run_potato_peel_with_custom_tmp` to try to fix that as well. Should be fixed in latest commit

tests/test_potato.py
tjgalvin: I believe in some of the singularity runnr functions there are checks to ensure the container path actually exists. Perhaps they are not needed with your monkeypatch'ing

tjgalvin:
I am going to give this a whirl now, and will merge if it works. By 'if it works' here I mean when I run the workflow *without* specifying extra configuration settings.

PR #318, opened 2025-05-28
Add functionality to flagger to flag around sunset and sunrise

ErikOsinga:
This PR adds a new functionality 

`flint_flagger flagtwilight` to flag a measurement set around local sunset/sunrise. 

It does require an additional dependency to figure out when those times are: [astroplan](https://github.com/astropy/astroplan)

An example call would be 

```
flint_flagger flagtwilight myms.ms -window 1800 --which nearest
```

which will flag a window of total width `2*1800` seconds around the nearest sunset and sunrise time in the MS (should they overlap).

### DIFF
--- a/flint/flagging.py
+++ b/flint/flagging.py
@@ -35,6 +35,7 @@ class AOFlaggerCommand(NamedTuple):
     strategy_file: Path | None = None
     """The path to the aoflagging strategy file to use"""
 
+
 class FlaggingOptions(BaseOptions):
     """Options provided to the flint_flagger module.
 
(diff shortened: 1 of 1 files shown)


ErikOsinga:
Now with 4 parameters to determine the window function, the call would be, e.g.

```
flint_flagger flagtwilight myms.ms \
  --pre-sunrise 360 \
  --post-sunrise 1800 \
  --pre-sunset  1800 \
  --post-sunset 360 \
  --which nearest
```

ErikOsinga:
I think it'd be nice to have the blanket flagging around sunrise and sunset as an option, although it might be superseded by the elevation flagging. 

I'm not sure what the best place is, I could add the functionality to the 
`jolly_roger` package instead, and then make it configurable in `flint` by adding to the `jolly` branch on this repository? 

Happy to take suggestions

ErikOsinga:
That was implemented into `flint` via the `flint_flagger` directly in this PR. I'm happy to close this and instead implement it as a possibility in `jolly_roger` instead. I can look into whether `astroplan` is absolutely necessary, but I'm on vacation for a week starting tomorrow, so it'll be a few weeks before I get to that.

tjgalvin:
Perfect. I am also going to be away a little.

I think parameterization as elevation makes it easy asier. And most of the
bits and pieces are not n jolly already. But I will have a look at this pr.
Flagging by elevation is something all wrapped up to you n astropy already,
so it would be easy to implement and n flint.

AlecThomson:
Going to close this for now in lieu of `jolly-roger`

PR #361, opened 2026-03-02
Test more python

AlecThomson:
Test PR to test more python versions. With tests.


AlecThomson:
Ok, Python 3.14 will still not work whilst we're on Prefect v2 (looks like Pydantic and Prefect don't play nice in that space)

PR #368, opened 2026-03-15
Switch to capn crunch

AlecThomson:


tjgalvin (approved):
All looks reasonable to me. Thanks for making this happen so quickly!

PR #374, opened 2026-04-25
Early RACS All pipeline

tjgalvin:
This PR is starting to get a little large and out of date from main. So now that it is in an 'image comes out the other end' it might be a good idea to merge back into main and continue iteration from there. 

The basic gist is there there is a separate RACS All workflow where the inputs are the three SBIDs of a matching field. The beam positions are expected to the same. The workflow will group related MSs, and a single wsclean command is issued for those three MSs. The nuts and bolts of the workflow itself is similar to the existing self-cal workflow. However, I have added some extensions to the MSSummary and FieldSummary objects and creation functions to attach existing summary objects rather than internally creating them. This allows a much more straight forward workflow with fewer instances of the `wait_for` parameter of `.submit()` and `.map()` being used. 

The RACS All workflow also separates out the per-beam imaging from the field convolution and linmosing a little more. Since the `wait_for`s have largely been factored out the workers generally are longer lived - once allocated there is enough work available to sustain them to the end of the flow. 

Around wsclean - I have added a `MSs` class. It started off as a type alias for a `list[MS]`, but it seemed like a good place to put consistency and sanity checks, so became a little more first class. I have extended the `MS.cast()` to return a `MS` is a list/tuple are provided as inputs. We should probably `@overload` the signature as the `MSs` is only returned on `list` or `tuple` usage.


AlecThomson:
Close in favour of #375 ?

tjgalvin:
Closed in favour of #375

PR #375, opened 2026-05-04
Early RACS-All pipeline - no MSs class

tjgalvin:
This PR is starting to get a little large and out of date from main. So now that it is in an 'image comes out the other end' it might be a good idea to merge back into main and continue iteration from there. 

This is more or less a duplicate of #374, with the exception that I have removed the `MSs` class, which was used to represent multiple separate instances of `MS`. In place of this I am using a simple `list[MS]`, and removed some of the checks that the `MSs` was responsible for, like ensuring all registered columns were consistent. So far the `wsclean` related code is the only operation that could process multiple MSs objects in a single call, so the switch was fairly straight forward. All internal functions in `flint.imager.wsclean` now operate against `list[MS]` after an initial conversion. 

The basic gist is there there is a separate RACS All workflow where the inputs are the three SBIDs of a matching field. The beam positions are expected to the same. The workflow will group related MSs, and a single wsclean command is issued for those three MSs. The nuts and bolts of the workflow itself is similar to the existing self-cal workflow. However, I have added some extensions to the MSSummary and FieldSummary objects and creation functions to attach existing summary objects rather than internally creating them. This allows a much more straight forward workflow with fewer instances of the wait_for parameter of .submit() and .map() being used. These `wait_for`s were required to prohibit the MS changing name (or being zipped) before the summary tool could run. 

The RACS All workflow also separates out the per-beam imaging from the field convolution and linmosing a little more. Since the wait_fors have largely been factored out the workers generally are longer lived - once allocated there is enough work available to sustain them to the end of the flow. The `wait_for`s are used between the imaging and self-cal tasks, largely to avoid any need to inspect the type of `WSCleanResult.ms` future. I have added a `mss` attribute to `WSCleanResult` now of type `list[MS]`, and in the case of a single `MS` `WSCleanResult.ms` and `WSCleanResult.mss` contain the same information, except the latter is in a list container.. 

Thoughts are welcome


flint/data/aoflagger/ASKAP.lua
```diff
@@ -28,7 +28,7 @@ that is packaged in aoflagger. The main differences are:
    -- If the following variable is true, the strategy will consider existing flags
    -- as bad data. It will exclude flagged data from detection, and make sure that any existing
    -- flags on input will be flagged on output. If set to false, existing flags are ignored.
-   local exclude_original_flags = true
+   local exclude_original_flags = false
```
AlecThomson, line 31: I'm glad to see this back TBH

flint/imager/wsclean.py
```diff
@@ -814,7 +828,7 @@ def _resolve_wsclean_key_value_to_cli_str(key: str, value: Any) -> ResolvedCLIRe
 
 
 def create_wsclean_cmd(
-    ms: MS,
+    ms: list[MS],
```
AlecThomson: Name as `ms_list` to be explicit?

flint/prefect/flows/racs_all_continuum_selfcal.py
```diff
@@ -0,0 +1,533 @@
...
+        output_split_science_path=output_science_path,
+        imaging_strategy=racs_all_options.imaging_strategy,
+    )
+    logger.info(f"Remove this later {strategy=}")
+
+    # Ya sea dog, we will only be handling CASDA measurementsets for the moment.
+    # We will consider bandpass applications later
+    _ensure_all_casda_format(mss_by_beams=science_mss_by_beam)
+
+    ms_summaries = []
```
AlecThomson: Type list?

flint/prefect/flows/racs_all_continuum_selfcal.py
```diff
@@ -0,0 +1,533 @@
...
+            )
+            imaging_results[0].append(
+                LoopFutures(
+                    mss=preprocess_science_mss,
+                    wsclean_result=wsclean_result,
+                    ms_summaries=ms_summaries_for_beam,
+                )
+            )
+
+    beam_summaries = []
```
AlecThomson: Add type?

flint/flagging.py
```diff
@@ -246,6 +246,11 @@ def create_aoflagger_cmd(ms: MS) -> AOFlaggerCommand:
     )
     logger.info(f"Flagging using the strategy file {flagging_strategy}")
 
+    # This is purely a temporary sanity test
+    with open(flagging_strategy) as strategy:
```
AlecThomson: Still needed?

flint/ms.py
```diff
@@ -317,14 +296,53 @@ def get_pol_axis_from_ms(
     return pol_ang
 
 
+def get_pol_axis_as_rad(ms: MS | Path) -> float:
```
AlecThomson, line 300: Can't get this from fixms?

AlecThomson (commented):
This is a beefy boy, so I can't promise I've read the entire PR deeply. I think we should still get it merged.

My only hold out would be places that have:
```
ms: MS | list[MS]
```
Would it be easier to just bite the bullet and always enforce `list[MS]`?

flint/flagging.py
tjgalvin: Delete, delete, delete, deleteeeee

flint/data/aoflagger/ASKAP.lua
tjgalvin: Also turns out I dont think we were using the flag_nans and flag_zeros functions properly. This has been fixed as well

flint/ms.py
tjgalvin: Could probably do that, yes. I'll do it next time I touch this.

flint/prefect/flows/racs_all_continuum_selfcal.py
tjgalvin: Yes

AlecThomson (approved):

PR #376, opened 2026-05-07
Add utility to concatenate holography cubes togethher

tjgalvin:
A fun and annoying one. With an eye to processing RACS-All I thought it would make sense to concatenated the three holography cubes required for a single field together to make processing 'simpler'. At the moment if we choose to linmos together cubes formed from the three bands we would have to provide a single cube that comprises all bands. 

Whether this is ultimately what we choose to do is uncertain. However, in this PR we have an appropriate routine with corresponding CLI entry point to perform the concatenation. Since between low/mid and high bands there are different grid spacings a reprojection also has to be performed before the concatenation takes place. 

This particular implementation uses a `ProcessPoolExecutor` in conjunction with a memory mapped numpy array to efficiently perform the reprojection and write out. We also use the same trick from `fitscube` wherein we write out the FITS file outside of `astropy.io.fits` by dumping the header to a file and then performing a further seek/write operation to write out a correctly sized file that can then be memory mapped. 

If one looks through the git history here you will see the different implementations that got to this point. All in we have gone from a 50 minute operation to 2ish minutes. The memory footprint is also much nicer as we don't really read in any of the FITS cubes into memory (in complete form at least).


flint/misc/holo.py
```diff
@@ -0,0 +1,582 @@
...
+
+@dataclass
+class FITSCubeInfo:
+    """Slim container to represent a FITS cube"""
+
+    path: Path
+    """Path to the cube being considered"""
+    header: fits.Header
+    """Header from the FITS cube"""
+    # data: NDArray[np.floating]
```
AlecThomson: Delete?

flint/misc/holo.py
```diff
@@ -0,0 +1,582 @@
...
+    # data = hdu[0].data
+    freqs_hz = get_freq_axis(header=header)
+
+    spatial_shape = (header["NAXIS2"], header["NAXIS1"])
+    celestial_wcs = celestial_wcs_from_header(header)
+
+    return FITSCubeInfo(
+        path=cube_path,
+        header=header,
+        # data=data,
```
AlecThomson: Delete?

flint/misc/holo.py
```diff
@@ -0,0 +1,582 @@
...
+    return in_header_2d
+
+
+@dataclass
+class ReprojectWorker:
+    in_cube_path: Path
+    """The cube to load"""
+    in_ch_idx: int
+    """The channel index of the input cube"""
+    # plane: NDArray[np.floating]
```
AlecThomson: Delete?

flint/misc/holo.py
```diff
@@ -0,0 +1,582 @@
...
+        mode="r+",
+        shape=dest_cube_info.output_shape,
+    )
+    arr[
+        reproject_worker.beam,
+        reproject_worker.stokes,
+        reproject_worker.out_idx,
+        :,
+        :,
+    ] = reprojected.astype(">f4")
```
AlecThomson, line 302: Will this _always_ be the correct data type?

flint/misc/holo.py
AlecThomson, line 1: May need to grab some of the float type stuff from fitscube just to be safe...

AlecThomson (changes requested):

flint/misc/holo.py
tjgalvin: Yes

flint/misc/holo.py
tjgalvin: Yes

flint/misc/holo.py
tjgalvin: Yes

flint/misc/holo.py
tjgalvin: In the initial creation it is hard coded for a floast32 size, which I think comes from the data type of the holography cubes themselves.

flint/misc/holo.py
tjgalvin: I will look into it. In my mind since the holography cubes are float32s hardcoding the size of the cube at the when making the placeholder cube and forcing a `astype(">f4")` should, at worst, lose prcesision if the holography cube type changes.

PR #377, opened 2026-05-09
Holography cube creation in the RACS-All pipeline

tjgalvin:
This PR adds a task to create a concatenated holography cube to the RACS-All pipeline, formed from the low-, mid- and high-band cubes. 

This is largely a straight forward PR, which the exception that I have moved away from `concurrent.futures.ProcessPoolExecutor` to `billiard.Pool`. The reason behind this was the when a `dask-worker` starts with a corresponding `nanny` process, the worker is set up in `daemon` mode and is unable to spawn child processes. The `billiard` package has its own implementation of a `Pool` that does not have such a limitation. 

The `billiard` package itself is supported by the `celery` package/team and its pretty popular. For the moment I'd like to move forward with this, and in the future we can port it to a `dask.delayed` approach should this become a major problem or we want to lower the dependency footprint.

### DIFF
--- a/flint/imager/wsclean.py
+++ b/flint/imager/wsclean.py
@@ -111,6 +111,8 @@ class WSCleanOptions(BaseOptions):
     """Threshold in Jy to stop cleaning"""
     channels_out: int = 4
     """Number of output channels"""
+    gain: float = 0.1
+    """Cleaning gain, ratio of peak that will be subtracted in each iteration"""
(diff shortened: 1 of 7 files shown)


tjgalvin:
I also renamed `ConcatHolo` to `ConcatHoloOptions` to be consistent with other aspects of the flow, and updated an return type hint.

PR #383, opened 2026-07-17
Better handling of the casda bandpass table flagging and carrying per-baseline channel flagging forward

tjgalvin:
For a long time we have configured aoflagger to reset the flag column when flagging visibility data. In the past any technical issues around the instrument, like bad intervals, missing antenna etc the data in the measurement set would have some other indicator or 'badness', including:
- (u,v,w) = (0,0,0)
- visibilities zero'd
- combinations of the two

So once aoflagger runs, in which is sets flags, we would always flag data based on the criteria above.

Since the usage of reference field calibration I have noticed this procedure can break. I believe there are situations where (for example) an antenna may be unavailable during the observation of the reference field, so bandpass table would flag visibility data when applying the solutions. If said antenna then became available in science observations, there are now visibilities that look real enough so wont be flagged by aoflagger, but they are uncalibrated. Inspecting the bandpass table applied to the data is one way of identifying such instances and re-flagging.

This PR does two thing:

    makes the examination of the casda supplied bandpass table more robust in how it searches and applies the VALID column examines the visibilities per-baseline directly to identeify instances where channels are flagged across all timesteps, and carries these flags forward after aoflagger.

### DIFF
--- a/flint/misc/interopt.py
+++ b/flint/misc/interopt.py
@@ -70,18 +70,19 @@ def get_flagged_antenna_casda_solutions(
         # the appropriate valid column
         flag_col_name = _find_casda_bandpass_table_type(columns)
 
-        # Remember that VALID is opposite to FLAG
-        flags = ~tab.getcol(flag_col_name)
+        # Remember that VALID is opposite to FLAG, so calling
+        # mask to try to highlight difference
+        mask = ~tab.getcol(flag_col_name)
 
-        # Flag shape will be (TIME, BEAM, ANT, GAIN).
-        assert flags.shape[0] == 1, f"More then one TIME interval in {flags.shape=}"
-        assert len(flags.shape) == 4, (
-            f"Expected dimensionality of 4, but got {flags.shape=}"
+        # Flag shape will be (TIME, BEAM, ANT, JONES_ELEMENT).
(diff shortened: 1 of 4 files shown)


flint/misc/interopt.py
```diff
@@ -35,15 +64,14 @@ def get_flagged_antenna_casda_solutions(
...
+        # the appropriate valid column
+        flag_col_name = _find_casda_bandpass_table_type(columns)
 
-        if not all(col in columns for col in expected_columns):
-            msg = f"{expected_columns=}, but found {columns=}"
-            raise ValueError(msg)
-
-        flags = ~tab.getcol("BANDPASS_VALID")
+        # Remember that VALID is opposite to FLAG
+        flags = ~tab.getcol(flag_col_name)
```
AlecThomson: Maybe rename as `mask` to avoid confusion with flags?

AlecThomson (approved):
All looking good to me. My only thought is just on the naming of flags vs masks. I've approved anyway since I don't think it matters. But it may be useful in the future.

flint/misc/interopt.py
tjgalvin: This makes sense for future generations

AlecThomson (approved):

PR #385, opened 2026-07-22
Parallel cube linmos

AlecThomson:
Fixes an outstanding `TODO` in the polarisation pipeline - we were previously making cubes and then giving to linmos. We were therefore missing an opportunity to parallelise by instead linmos-ing images then forming a cube from the mosicked channels.

Old: per beam → cube → one linmos per stokes crunching all ~288 channels in a single serial process.

New:

1. Per beam: split → convolve → keep per-channel images (no per-beam cube).
2. Transpose beam×channel → channel×beam (transpose_and_sort_channel_images, sorted by channel range).
3. One linmos per channel across beams — fanned out in parallel via `.submit()`.
4. Stack per-channel mosaics into image + weight cubes.
5. Cleanup: per-channel weight files (`LinmosOptions.cleanup=True`), mosaics (`remove_original_images=True` on the cube step), and convolved beam images (final task_remove_files_folders, gated on all cubes).
6. Correctness preserved: Q/U leakage correction now passes the matching per-channel Stokes-I beam images as `stokesi_images`; `force_remove_leakage=False` when I isn't imaged; Stokes-I inputs kept until Q/U consume them. A `.result()` barrier after convolution guarantees all files exist on disk before any linmos runs.

New code: `transpose_and_sort_channel_images` in `wsclean.py` (+ test), and three thin `task(...)` wrappers in `imaging.py`. Dropped the now-unneeded `task_rename_linear_to_stokes`/`task_create_object` hack — names derive from the actual per-stokes files.

### DIFF
--- a/flint/imager/wsclean.py
+++ b/flint/imager/wsclean.py
@@ -42,6 +42,7 @@ from flint.exceptions import (
 from flint.logging import logger
 from flint.ms import MS, standardise_ms_to_list_ms
 from flint.naming import (
+    ProcessedNameComponents,
     create_image_cube_name,
     create_imaging_name_prefix,
     extract_components_from_name,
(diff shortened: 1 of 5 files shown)


tjgalvin:
This looks good to me. I have not been able to run any actual workflow instances, but it all seems to track. A big thanks for pulling this one together.

tjgalvin (approved):

PR #386, opened 2026-07-24
Rearrange the steps in CASA gaincal/applycal, potentially avoid crashes on restart of task

tjgalvin:
This PR is trying to work towards a fix for a intermittent problem. Occasionally during a self-calibration flow the some `task_gaincal_applycal` task can restart. As far as I can tell, something temporary happens to a worker, and the task is re-started. 

During normal flow of this task there are some side effects, like the renaming of the MS, deleting columns, etc. On the task restarting these side-effects are not accounted for, and the house of cards comes tumbling down. 

I have tried to add some components to handle this a little more gracefully. The typical situation I use nowadays is to simply rename the MS (avoiding duplicating the MS) into the new named path, and then deleting DATA and renaming CORRECTED_DATA to DATA. Should this appear to be the case (old MS name does not exist but the next one does, and the columns match what I've described above) the function is allowed to continue. 

I have also added a check for the output calibration table. if it already exists, that specific gaincal is skipped. I have also made it so an error is always raised if the calibration table does not exist at this point, where as previously there was an optional keyword that would suppress the exception. 

Finally that has been a slight tweak to the application of the solutions. I am now applyin them all sequentially after all solutions have been formed. Since these are transforming DATA to CORRECTED_DATA these can be repeated any number of times withoput changing the output. 

In doing this I think the code could be significantly reduced in complexity, which I will raise a separate issue for.


flint/selfcal/casa.py
```diff
@@ -52,6 +52,44 @@ class GainCalOptions(BaseOptions):
...
+    if not out_ms_path.exists():
+        msg = f"Neither {ms.path=} nor {out_ms_path=} exists. No attempt to recover."
+        raise ValueError(msg)
+
+    # Need to make sure that the DATA column exists for
+    # subsequent gaincal task
+    with table(str(out_ms_path), ack=False, readonly=True) as tab:
+        column_names: list[str] = tab.colnames()
+
+    return "DATA" in column_names
```
AlecThomson: **Recovery check can accept a half-renamed MS and silently lose a round of solutions.**

This only checks `"DATA" in column_names`, not that `CORRECTED_DATA` is absent. `rename_ms_and_columns_for_selfcal` renames first and *then* does `removecols("DATA")` + `renamecol("CORRECTED_DATA", "DATA")`. The comment above at line 79 treats the rename as atomic, but the rename+column-swap **pair** is not, and the window is long (`removecols` on a huge column).

A worker killed in that window leaves the output MS with both the previous round's uncalibrated `DATA` and its calibrated `CORRECTED_DATA`. Recovery then returns `True`, gaincal solves against the stale `DATA`, and applycal overwrites the good `CORRECTED_DATA` — one round of solutions silently lost, no error raised.

Suggest requiring `"DATA" in column_names and "CORRECTED_DATA" not in column_names`, and finishing the column swap otherwise.

flint/selfcal/casa.py
```diff
@@ -298,29 +371,32 @@ def gaincal_applycal_ms(
...
-            minsnr=gain_cal_options.minsnr,
-            calmode=gain_cal_options.calmode,
-            selectdata=gain_cal_options.selectdata,
-            uvrange=gain_cal_options.uvrange,
-            solnorm=gain_cal_options.solnorm,
-        )
+        # This check attempts to avoid reduplication of work if already
+        # carried out. Trying to make this function pure with no side
+        # effects, matie
+        if cal_table.exists():
```
AlecThomson: **Skip guard keys only off the caltable path, so it ignores changed gaincal options.**

`create_and_check_caltable_path` derives the path from the MS name and channel range only — never from the gaincal options. So a re-run after changing `solint`/`calmode`/`uvrange`/`minsnr` silently reuses the old solutions. Previously gaincal overwrote the table (`append` unset, CASA default `False`), so this is a behaviour change.

`create_and_check_caltable_path` already has a `remove_if_exists` flag that is currently unused — plumbing that through from the caller would give an explicit way to force a re-solve.

flint/selfcal/casa.py
```diff
@@ -298,29 +371,32 @@ def gaincal_applycal_ms(
...
+            logger.warning(f"Found an earlier {cal_table=}. Not rerunning gaincal!")
```
AlecThomson, line 425: **`exists()` is too weak a completeness test for a caltable.**

It is true the moment CASA creates the table directory, so a table left half-written by a killed gaincal is accepted and handed straight to applycal. gaincal is the longest step and hence the most likely restart point — precisely the case this PR targets.

Needs at least an "opens and has rows" check, e.g. open with `casacore.tables.table` and assert `nrows() > 0` before deciding to skip.

AlecThomson (commented):
NOTE: This is a Claude review (was experimenting)
Code review of the restart-recovery changes. Five findings inline, most severe first.

Not filed as findings:
- The `interopt.py` `np.atleast_1d` fix is correct: `squeeze` returns a 0-d array when exactly one antenna is flagged, which would blow up `len()` and the `for` loop in `flag_antenna_from_casda_bandpass_table`.
- Removing `raise_error_on_fail` breaks no in-repo caller.
- The archiving loop leaves `idx`/`spw_str` unused, and its comment is garbled mid-sentence ("If we ought to make sure...").
- The new tests use the function-scoped `ms_example` fixture, so the in-test `rename` is safe. I could not execute them locally (no pytest in the project venv).

flint/selfcal/casa.py
```diff
@@ -80,20 +118,25 @@ def copy_and_clean_ms_casagain(
...
-        logger.warning(f"{ms.path=} {ms.path.exists()=}")
-        logger.warning(f"{out_ms_path=} {out_ms_path.exists()=}")
-        remove_files_folders(out_ms_path)
+    if _skippable_ms_copy_and_clean(
+        ms=ms, out_ms_path=out_ms_path, rename_ms=rename_ms
+    ):
+        ms = ms.with_options(path=out_ms_path, column="DATA")
 
-    if rename_ms:
+    elif rename_ms:
```
AlecThomson, line 173: **Moving the stale-output cleanup into the copy-only `else` makes the rename branch raise `FileExistsError`.**

The `out_ms_path.exists()` -> `remove_files_folders` guard now lives only in the `else` at line 134, so the rename path no longer clears a stale output and instead hits the `FileExistsError` in `flint/ms.py` (`rename_ms_and_columns_for_selfcal`).

Reachable by re-running over a directory left behind by a `rename_ms=False` run with `rename_ms=True`.

flint/selfcal/casa.py
```diff
@@ -52,6 +52,44 @@ class GainCalOptions(BaseOptions):
     """Renormalise the distribution of the amplite gains so the mean is unity."""
 
 
+def _skippable_ms_copy_and_clean(
+    ms: MS, out_ms_path: Path, verify: bool = True, rename_ms: bool = False
```
AlecThomson: `verify` is declared but never read and never passed by the caller; the docstring also omits it and omits the `ValueError` raise. Suggest dropping the parameter and adding a `Raises:` section.

flint/selfcal/casa.py
tjgalvin: I think I agree with this. Extended the check here.

flint/selfcal/casa.py
tjgalvin: I dont expect to have to support this, yet at least. If the task is rerun then the same options are going to be used.

flint/selfcal/casa.py
tjgalvin: agree

flint/selfcal/casa.py
tjgalvin: This won't happen as I don't expect to rerun the task with different input options

AlecThomson (approved):

PR #388, opened 2026-07-25
Parallel cube linmos in the self-cal pipeline

AlecThomson:
`linmos` co-adds a cube one channel at a time, serially, so co-adding the
per-beam continuum cubes was a single long-running task at the end of the
pipeline. #385 fixed this for the polarisation flow by co-adding each channel in
parallel and stacking the resulting mosaics into a cube. This does the same for
the continuum (self-cal) flow.

The wrinkle is that the continuum flow has already combined each beam's sub-band
images into a cube by the time the co-add happens (the sub-band images are
deleted at that point), so the beam cubes are split back into per-channel planes
before the co-add:

- `split_cube_into_planes` extracts each channel of a cube into its own image,
  named with a `chNNNN-NNNN` field so the existing
  `transpose_and_sort_channel_images` can regroup the planes across beams.
- `linmos_channel_groups_to_cubes` runs a `linmos` per channel and stacks the
  mosaics into the image and weight cubes. This is now shared by both the
  continuum and polarisation flows, replacing the loop that #385 added inline to
  the polarisation flow.

The extracted planes and the per-channel mosaics are removed once the cubes are
built.

### Note on output names

As in the polarisation flow, the co-added cubes are now named
`<prefix>.image.cube.fits` and `<prefix>.weight.cube.fits`, rather than the
single `<prefix>.linmos.fits` that `linmos` produced directly.

### Tests

`test_split_cube_into_planes` builds a cube from the packaged sub-band images,
splits it, and checks the planes are named such that they still regroup per
channel.


tjgalvin (approved):
A nice change, and I am glad we figured out the trimming issue. 

In time it might be worth splitting out some of the ImageSet operations from wsclean namespace into its own, and with that bring forward some of the cube splitting/joining operations. The latter may only be worth it if they are generic in nature - if they are tied fundamentally to wsclean outputs than we don't need to. 

Very happy with how this performed in practise, and will put it into the racs-all pipeline soon.

PR #389, opened 2026-08-02
Prefect3 migratation

AlecThomson:
Migrates to Prefect v3
- Catches changes in v2-v3, notably task result caching

Still todo - run some smoke tests on real data with a flow


tjgalvin (approved):
I am happy :D. 

I have put the majority of the flows through their paces and all seems to be working. Any issues we will deal with.

PR #391, opened 2026-08-05
Add RM synthesis

AlecThomson:
For profit


AlecThomson:
Will merge racs-all-pol in here, and make this downstream of that

AlecThomson:
So we need to merge #390 *before* this one

PR #393, opened 2026-08-06
Refactor cube convolution

AlecThomson:
Refactors how per-cube cubes are made during the self-cal flows. Currently we convolve the 3D cubes which can blow up memory (annoyingly).

This changes the ordering to `split into planes → group by channel across beams → convolve each plane (2D) → linmos`


tjgalvin (approved):
This works nicely. Also, the switching of the ordering (first continuum linmos and then the linmos csubes) also worked as expected. Don't want to over thinking it but in my before/after test we were 30minutes faster as workers were more continually saturated with tasks.

PR #397, opened 2026-08-08
Add fitscube options

AlecThomson:


tjgalvin (approved):
Tested this and it all works, thanks for putting this one together. 

Note I did have to switch the check and overwritting of the compress option on the last round of imaging.

PR #403, opened 2026-08-17
RACS all pol refactor

AlecThomson:


flint/prefect/common/spice.py
```diff
@@ -23,9 +23,7 @@ def task_get_spice_boxes(
     beam_shape: BeamShape,
     is_user_catalogue: bool,
 ) -> list[BoundingBox]:
-    """Build the (single, shared) set of island bounding boxes for a field,
-    against ``reference_image``'s WCS/shape -- reused as-is for every
-    Stokes/plane/cube ``task_spice_fits`` is later applied to."""
+    """Build the (single, shared) set of island bounding boxes for a field"""
```
AlecThomson:
```suggestion
    """Build the set of island bounding boxes for a field"""
```

flint/options.py
```diff
@@ -351,6 +339,49 @@ class SpiceOptions(BaseOptions):
...
+    stokes_u_cube: Path
+    """Path to the Stokes U FITS cube"""
+    stokes_i_cube: Path | None = None
+    """Path to a Stokes I FITS cube, used to fit a per-pixel fractional-polarisation correction. Defaults to None."""
+    imaging_strategy: Path | None = None
+    """Path to a FLINT imaging yaml file that contains the RMSynthOptions/RMCleanOptions settings to use"""
+    cube_products: list[Literal["dirty", "clean", "model"]] = []
+    """Which Faraday dispersion function (FDF) cubes to write as FITS. 'dirty' is the raw FDF (no RM-CLEAN needed). 'clean'/'model' are RM-CLEAN's cleaned/clean-component FDF (RM-CLEAN is run if either is requested). Empty by default, as these cubes are large."""
+    moment_products: list[Literal["dirty", "clean", "model"]] = ["clean"]
+    """Which FDF(s) to compute Faraday moment maps (mom0=polarised intensity, mom1=mean Faraday depth, mom2=Faraday depth dispersion) from. 'clean' is the usual choice; 'dirty' moments are noise-biased; 'model' moments describe just the clean components."""
```
```suggestion
    """Which FDF(s) to compute Faraday moment maps from."""
```

tests/test_prefect_racsall_flow.py
```diff
@@ -13,3 +13,19 @@ def test_get_parser() -> None:
     parser = get_parser()
 
     assert isinstance(parser, ArgumentParser)
+
+
+def test_get_parser_has_no_run_polarisation() -> None:
```
tjgalvin: I am not sure if we need this test. Seems a little backwards facing.

flint/options.py
```diff
@@ -351,6 +339,49 @@ class SpiceOptions(BaseOptions):
...
+    stokes_q_cube: Path
+    """Path to the Stokes Q FITS cube"""
+    stokes_u_cube: Path
+    """Path to the Stokes U FITS cube"""
+    stokes_i_cube: Path | None = None
+    """Path to a Stokes I FITS cube, used to fit a per-pixel fractional-polarisation correction. Defaults to None."""
+    imaging_strategy: Path | None = None
+    """Path to a FLINT imaging yaml file that contains the RMSynthOptions/RMCleanOptions settings to use"""
+    cube_products: list[Literal["dirty", "clean", "model"]] = []
+    """Which Faraday dispersion function (FDF) cubes to write as FITS. 'dirty' is the raw FDF (no RM-CLEAN needed). 'clean'/'model' are RM-CLEAN's cleaned/clean-component FDF (RM-CLEAN is run if either is requested). Empty by default, as these cubes are large."""
```
```suggestion
    """Which Faraday dispersion function (FDF) cubes to write as FITS. Nothing by default, as these cubes can be large."""
```

flint/options.py
```diff
@@ -351,6 +339,49 @@ class SpiceOptions(BaseOptions):
...
+class SpiceFieldOptions(BaseOptions):
+    """Options for running SPICE-style cube compression as its own standalone
+    pipeline, given already-imaged Stokes cubes. See ``SpiceOptions`` for the
+    trimming/compression algorithm parameters, which are drawn from
+    ``imaging_strategy`` rather than exposed here."""
+
+    cubes: list[Path]
+    """Stokes cubes to trim and compress"""
+    reference_image: Path | None = None
+    """A 2D MFS image whose WCS/shape sources the source-finding boxes. Required only when catalogue is not set (built-in aegean source finding); a supplied catalogue instead sources WCS/shape from one of cubes' own header"""
```
```suggestion
    """A 2D MFS image whose WCS/shape sources the source-finding boxes. Required only when catalogue is not set (built-in aegean source finding)"""
```

flint/prefect/common/rmsynth.py
```diff
@@ -13,37 +13,70 @@
...
+    run_rmsynth_3d,
+    write_rm_products,
+)
 
 
 @task
-def task_rmsynth_and_write_products(
+def task_rmsynth(
     stokes_q_cube: Path,
     stokes_u_cube: Path,
```
tjgalvin, line 29: My preference would be to:

```python
task_rmsyth = task(rmsynth)
```

if the options are all one-to-one. Otherwise, if we add an option to the function rmsynth we now need to update the function defition of another task decorated function.

flint/prefect/common/rmsynth.py
```diff
@@ -13,37 +13,70 @@
...
+    enable_rmlite_logging_support()
+    return run_rmsynth_3d(
+        stokes_q_cube=stokes_q_cube,
+        stokes_u_cube=stokes_u_cube,
+        rmsynth_options=rmsynth_options,
+        stokes_i_cube=stokes_i_cube,
+    )
+
+
+@task
```
tjgalvin, line 48: Same here. 

The exception being if there is something we want to wrangle in the task function (e.g. extracting an attribute from a class we don't want in the scope of the wrapped function), but it does not seem like that this is an issue

flint/prefect/flows/racs_all_pipeline.py
```diff
@@ -0,0 +1,295 @@
...
+    _check_racs_all_pipeline_options(pipeline_options=pipeline_options)
+    _check_spice_mfs_dependency(
+        pipeline_options=pipeline_options,
+        racs_all_options=racs_all_options,
+        spice_field_options=spice_field_options,
+    )
+
+    terminal_results: list[Any] = []
+
+    if not pipeline_options.run_imaging:
```
tjgalvin: I am a little unsure if this check is needed.

tjgalvin: Would it be better to flip the if condition around:

```pythong
if run_image:
   imaging_flow.with_options(...)(...)
```

Not sure what this check is really trying to do. Should we decide not to run the imager then the whole flow quits?

flint/prefect/flows/spice_compression_pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+
+    island_boxes = task_get_spice_boxes.submit(
+        reference_image=reference_image,
+        catalogue=catalogue_path,
+        spice_options=spice_options,
+        beam_shape=beam_shape,
+        is_user_catalogue=is_user_catalogue,
+    )
+
+    spiced_cubes = [
```
tjgalvin: Would the `.map` be better python here?

```
task_sice_fits.map(fits_path=spice_field_options.cubes, boxes=unmapped(island_boxes))
```

flint/prefect/flows/spice_compression_pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+        task_spice_fits.submit(fits_path=cube, boxes=island_boxes)
+        for cube in spice_field_options.cubes
+    ]
+
+    compressed_cubes = [
```
tjgalvin: Same `.map` usage here as well

flint/prefect/helpers.py
```diff
@@ -66,3 +68,39 @@ def enable_loguru_support() -> None:
...
+        logging.ERROR: "error",
+        logging.CRITICAL: "critical",
+    }
+
+    def emit(self, record: logging.LogRecord) -> None:
+        method_name = self._LEVEL_TO_METHOD.get(record.levelno, "info")
+        getattr(get_run_logger(), method_name)(self.format(record))
+
+
+def enable_rmlite_logging_support() -> None:
```
tjgalvin: Teo thoughts. 
1 - I would prefer we generalise away from rm-lite and make it general to any module. I am already using `PREFECT_LOGGING_EXTRA_LOGGERS` to record logs from other modules, including fixms, fitscube and flint itself. This goes into the next point. 
2 - Todate it has been useful to not have to remember to do any `logger = get_run_logger()` prefect magic within every task. Attaching the prefect log handler on start up with the existing prefect infrastructure has made this seemless. If we can run this registation function at the start of every flow so that all out desired modules are captured across all tasks, great. Be nice to include. If not and we need to run it at the task level, my worry is it will get clunky to remember, particularly along side the existing approach. Of course, if this is doing something that the `PREFECT_LOGGING_EXTRA_LOGGERS` is not doing (e.g. making debug look like info, such as what I think may been needed for fitscube) then I am a little more comfortable.

flint/options.py
```diff
@@ -459,6 +486,61 @@ class or through ``RACSAllOptions`` itself for the shared ones."""
...
+    it rediscover measurement sets and paths by globbing the output
+    directory."""
+
+    mss_by_beam: tuple[tuple[MS, ...], ...]
+    """The final round's per-beam self-calibrated measurement sets. Structurally equivalent to ``flint.ms.MSsByBeam``"""
+    holography_path: Path | None
+    """Path to the holography FITS cube used when co-adding beams, if any"""
+    output_science_path: Path
+    """Directory the continuum imaging stage wrote its output products into"""
+    terminal_futures: list[Any]
```
tjgalvin: I don't want to intermix perfect specific things with the larger code base. We should try to refactor this out.

flint/options.py
```diff
@@ -459,6 +486,61 @@ class or through ``RACSAllOptions`` itself for the shared ones."""
...
+class PolPipelineResult(BaseOptions):
+    """Return value of ``process_science_fields_pol``, handed in-memory to
+    the rm-synth/clean and spice-compression stages of the ``racs-all``
+    flow-of-flows."""
+
+    stokes_cubes: dict[str, Path]
+    """The full, unspiced Stokes cube written for each imaged polarisation (e.g. 'i', 'q', 'u', 'v')"""
+    mfs_products: dict[str, dict[str, Path]]
+    """MFS image/model/residual products co-added per Stokes parameter, keyed by Stokes ('i', 'q', 'u', 'v') then product type ('image', 'model', 'residual'). Only populated for Stokes imaged under a polarisation with ``WSCleanOptions.flint_save_mfs_products`` set"""
+    terminal_futures: list[Any]
```
tjgalvin: Same as above. If it is prefect specific then we should consider trying to remove it.

PR #404, opened 2026-08-23
Make RM-synthesis fast and correct at RACS-all scale

AlecThomson:
## The short version

The RM-synthesis stage was spending almost all its time fitting a spectrum to
every pixel of empty sky, and producing infinities while it did so. On a test
cube built on the real RACS-all frequency coverage, this takes it from *never
finishing* to 8 seconds.

| | before | after |
|---|---|---|
| Stokes I fit + dirty moments | 407 s, `mom0 max = inf` | **6.0 s**, finite |
| the same plus RM-CLEAN | >900 s, never finished | **8.1 s** |

256×256 cube, 4-worker dask cluster. Scaling is linear from there: 512×512 in
37 s. Both numbers were measured against the *released* rm-lite, so they don't
depend on the companion branch below.

## Why it was slow

If you hand the pipeline a Stokes I cube, rm-lite fits a spectrum to each pixel,
divides Q/U by it, and works in fractional polarisation. Fitting *every* pixel
would be silly, since most of a field is empty sky, so there is a guard:
`stokes_i_snr_cut = 5.0`. Only pixels detected above 5 sigma get fitted.

To decide "is this above 5 sigma?" you need a noise value. rm-lite computes
`mean(I) * sqrt(n) / rms(error)`, and if you give it no error array it fills one
with zeros. That would divide by zero, so rm-lite guards it by returning
infinity — a sensible choice on its own, meaning "no noise information, so don't
reject anything".

But we passed no error array. `estimate_stokes_i_noise` defaulted to `False`, so
every pixel scored infinity, cleared the cut, and got a full curve fit. The
guard was armed with nothing to fire.

That is 24.7 ms per empty-sky pixel against 16.8 µs to reject one — roughly
1500x, on what is very nearly the whole cube.

## Why it was also wrong

This is the part that matters beyond speed.

A second-order power law has three free parameters. Fitted to a real source, the
data pins them down. Fitted to noise, nothing does, and the curve wanders. Of
400 noise pixels I fitted, 277 produced models that dip below 1e-6 somewhere in
the band, the median low point was 1e-10, and one hit exactly zero.

rm-lite then divides Q/U by that model. Dividing by 1e-10 multiplies the signal
by ten billion; dividing by zero gives infinity. The pipeline was emitting
`mom0 = inf`.

And that explains the *other* symptom. RM-CLEAN keeps cleaning while the signal
is above its threshold. An infinite signal is always above it, so RM-CLEAN ran
to its 100,000-iteration limit on those pixels. The slow RM-CLEAN was not a
separate problem — it was this one arriving downstream.

## The fix

`estimate_stokes_i_noise` now defaults to `True`, which derives a robust
per-channel noise from the Stokes I cube itself in one extra pass. With that,
none of 600 pure-noise pixels get fitted, and every model that *is* fitted is
well behaved.

Rejecting a pixel is safe, not approximate: it gets a flat model, and dividing
by a flat model then rescaling by it cancels exactly. I checked rather than
assumed — 64,270 of 65,536 pixels took that path and their output matched the
no-Stokes-I result to 3e-11.

And the broken pairing is now **refused rather than warned about**:
`RMSynthOptions` rejects `stokes_i_snr_cut` set alongside
`estimate_stokes_i_noise=False`. It is a pydantic validator, so it fires at
option construction — for a strategy file that means at load, before imaging
runs, rather than hours later — and `with_options` re-validates, so it survives
the racs-all flow rebuilding the options.

## Two other things this fixes

**The dirty Faraday moment maps had no noise cut.** `mom0` adds up the
polarised signal across all 373 Faraday depths, so with no cut an empty-sky
pixel accumulated hundreds of noise samples into a positive floor — a value of
4.7 where the actual noise level is 0.11. The first and second moments were then
weighted by that noise and meant nothing off-source.
`RMCleanOptions.moment_threshold_snr` now applies to every moment map flint
writes, the dirty ones included. Empty sky reads zero.

**RM-CLEAN was being run more than once.** Every output product is built from
one shared calculation per piece of the image, and we compute them all together
specifically so that calculation happens once. It didn't. Asking for two FDF
cubes ran RM-CLEAN twice, three cubes three times, one cube plus its moment maps
twice. The cause is that dask's writer optimises the recipe it captures at the
moment you build it, and in doing so copies the shared step into each output's
private recipe, so they no longer recognise each other as the same work. Fixed
by writing all the cubes through a single call with that optimisation switched
off for it. The default settings never hit this, since no FDF cubes are written
by default, but it was a trap in a documented option.

## Multiscale RM-CLEAN is not supported for now

I had exposed rm-lite's seven multiscale RM-CLEAN parameters here, having found
that only the `multiscale` bool was reaching rm-lite at all. Those went back out
first, and the switch itself has now followed: flint does not support multiscale
RM-CLEAN for now. It is experimental in rm-lite, an option flint ships is one it
has to keep meaning the same thing, and a switch flint cannot stand behind is
worth less than no switch.

Nothing changes behaviourally — the field defaulted to `False`, and with nothing
forwarded rm-lite applies its own `False`. What does change is that
`RMCleanOptions` forbids extra fields, so a strategy setting `multiscale` under
`rmclean` is now **refused at load rather than silently ignored**. That is the
point rather than a side effect: the failure mode it replaces is a run that looks
like it honoured the setting. A test pins it.

Adding it back is a one-line field plus one forwarded argument, plus a
`multiscale_`-prefixed field for each of the seven parameters rm-lite takes to
tune it.

Worth recording from the audit that prompted this, since it was done in both
directions: apart from those, every `RMCleanOptions` field reaches rm-lite.
`log_level` is the only rm-lite parameter deliberately left unset — its defaults
(WARNING for synthesis, ERROR for CLEAN) are what you want at cube scale, where
rm-lite otherwise logs per pixel. `debias_moments` and `debias_filter_size` are
not rm-lite arguments by design, as flint builds the moment maps itself. The
audit also turned up `MultiscaleOptions.selection` having no route through
rm-lite's 3D wrappers at all; that is fixed on the rm-lite branch, and would be
reachable from here if flint ever supports multiscale.

## On the rm-lite version

It briefly needed an unreleased rm-lite. `reuse_rmsf` on `rmsynth_3d_from_fits`
was the last such argument, and rm-lite #65 changed that API before releasing it:
the RMSF optimisation is now unconditional, so the flag is gone from the 3D entry
points and survives only on `get_rmsf_nufft` for testing. The flag is therefore
dropped here too, along with the `RMSynthOptions` field, which would otherwise
have been a strategy option reaching nothing. Nothing is lost by that — the
optimisation is on by default and needs no opt-in.

The pin is now `rm-lite==2026.8.2`, exact. rm-lite is pre-1.0 and versioned by
date, so it makes breaking API changes in ordinary releases, 2026.8.2 included,
and flint should only move when someone has read the diff. `pyproject.toml`
and `uv.lock` are the only two places flint declares it.

**2026.8.2 stopped shipping the 3D RMSF as a per-pixel cube.** A pixel's RMSF
depends only on which channels it has flagged, and flagging is per channel, so
`RMSynth3DResults.rmsf_arr` is now the one spectrum the whole cube shares.
Breaking, but free for flint to follow: flint never read `rmsf_cube`, never wrote
the RMSF out, and calls `run_rmclean_from_synth`, which picks the right form
itself. One docstring was stale. `per_pixel_rmsf=True` gets the old cube back and
is deliberately left off, with a comment saying why — it is ~2x the FDF's size
and would hold that one spectrum in every pixel.

## The fitted Stokes I model is now written out

2026.8.2 also returns the full fitted Stokes I model, not just alpha. flint
writes it as **one named FITS per term** rather than an anonymous
`(n_coeff, ny, nx)` cube, because the term names are what make it usable:
`stokesi.coeff.flux`, `.alpha`, `.beta`, ... for the log fit, `.c0`, `.c1`, ...
for the linear one, each with its `_error` companion.

The terms plus the reference frequency and the fit function *are* the whole
model, so anything downstream can evaluate Stokes I at any frequency from these
maps alone without carrying the model cube. That only works if the maps say which
term is which and what they are referenced to, so each header carries `REFFREQ`,
`FITFUNC`, `SICOEFF` and `SICOEFFI`. The existing `ref_flux`/`alpha`/`model_order`
maps get `REFFREQ` and `FITFUNC` too: a spectral index means nothing without the
frequency it is defined at, and they were being written without it.

Zero and NaN mean different things and both are recorded in the header. A zero is
the *actual value* of a term the AIC dropped — it contributes nothing to the
model, and `stokesi.model_order` says how many terms were really fitted — while
a NaN is a pixel that was never fitted at all.

Rather than assume the plane naming and ordering, a test rebuilds the model from
the written maps and their headers and checks it against the model cube rm-lite
fitted: agreement to 1e-5, i.e. float32. The same test pins the zero convention,
since a pure power law makes the AIC drop the higher terms.

**One thing this had to fix to be affordable.** These maps were being gathered
into the single `write_rm_products` process in float64 and written as float32.
Adding six more full-size planes that way would have taken what that one process
holds on a 16032^2 mosaic from ~8 GB to ~19 GB. They are now cast to float32
before the gather instead of after, which changes no written byte and halves the
peak.

## Testing

74 tests passing across the RM-synthesis, flow, configuration and options suites
against the released `rm-lite==2026.8.2`, full pre-commit clean including mypy.

New tests cover: the per-term Stokes I maps and their headers, under both fit
functions; the model rebuilding from those maps; a strategy asking for multiscale
RM-CLEAN being refused; the `estimate_stokes_i_noise`
default; the SNR-cut pairing being refused via the constructor, via
`with_options`, and via options loaded from a strategy file; a noise-only cube
coming back finite and empty; the dirty moment threshold actually applying; and
parametrised checks that RM-CLEAN runs exactly once per image piece, and that the
Stokes I fit count does not move, for every combination of requested products. I
confirmed both fail if the fix is reverted.

The flow-level cluster test now has a Stokes I case. Without one, nothing in the
fitted Stokes I path ran on a cluster: the per-term maps are sliced out of a
single `map_blocks` after the gather, and an unfitted run reports `None` for all
of them, so a pickling or plane-ordering fault would only have shown on the path
that fits.

**Those cluster tests now run in CI.** All three were marked `slow`, and CI runs
pytest with `--skip-slow`, so none of them executed there — the flow's prefect
and dask-distributed wiring was covered only by whoever ran the suite locally.
The marker is off. It costs about 80 s on a job that takes four and a half
minutes, and the three of them survive `-n auto`: three concurrent
`LocalCluster`s on four cores, clean across three runs.

Run end to end outside CI as well, on a 3-worker process cluster with a
288-channel 128x128 field: the flow completes in 23.6 s for the default product
set and 32.7 s asking for both FDF cubes and both moment sets, peak worker RSS
546 MiB, no restarts or spilling. `I(nu)` rebuilt from the written FITS maps and
their headers alone matched the injected curved power law to 2e-4, clean `mom1`
came back 50.00 +/- 0.05 against a true 50, and RM-CLEAN ran exactly 128 times
for 128 chunks with four products requested.

`docs/config.md` and `CHANGELOG.md` are untouched, per review.

## Companion rm-lite changes

Branch `claude/racs-all-clean-speed`. **None of these is required by this
branch** — they are independent improvements, and the pin can move whenever they
release:

1. **Skip RM-CLEAN on pixels where it provably has nothing to do.** Verified
   byte-identical output in all four modes; also fixes a crash on fully blank
   spectra in multiscale.
2. **Refuse a fitted Stokes I model too badly behaved to divide by** — the
   backstop for the bug above.
3. **Reuse one RMSF when the flagging is shared** — released in 2026.8.2, so
   this one is in the pin. Found from here: the RMSF is needed only by RM-CLEAN
   and depends only on a pixel's channel flagging, so a normal cube has one
   distinct RMSF per chunk, not one per pixel. On a
   2048-wide field chunk the per-pixel path cost 0.93 s and 233 MiB to produce
   one spectrum's worth of information. Now detected and collapsed
   unconditionally inside rm-lite, which is why flint needs no flag for it.
4. **Expose `MultiscaleOptions.selection`**, which had no route through the 3D
   wrappers.
5. **Share the synthesis graph with RM-CLEAN's copy of it.** `run_rmclean` takes
   the dirty cube through `to_delayed()`, whose dask default is
   `optimize_graph=True`, which optimises and so re-keys the whole upstream
   graph. Nothing computed off the synthesis results can then share it, so the
   Stokes I fit runs twice per chunk whenever the maps are written alongside
   clean products, and the NUFFT runs twice if both dirty and clean products are
   asked for. Passing `optimize_graph=False` there gives byte-identical results
   with all three kernels back to once per chunk — 21.1 s to 15.6 s on a field
   where every pixel is fitted. Flint needs no change for it; the guard
   described above pins that flint's own batching adds nothing on top.

Plus a fix for `write_zarr_group`, which had the same shared-work problem as the
second of the "two other things" above, and whose own test was already failing
on rm-lite `main` (84 RM-CLEAN passes for 21 chunks).


docs/config.md
AlecThomson, line 1: @claude - we won't update docs yet

flint/prefect/common/rmsynth.py
```diff
@@ -62,7 +67,6 @@ def task_write_rm_products(
     clean_results: RMClean3DResults | None,
     stokes_q_cube: Path,
     rmsynth_options: RMSynthOptions,
-    rmclean_options: RMCleanOptions,
```
AlecThomson, line 65: @claude - why did clean options get deleted?

flint/prefect/flows/rmsynth_pipeline.py
```diff
@@ -65,6 +65,7 @@ def process_rmsynth(rmsynth_field_options: RMSynthFieldOptions) -> list[Path]:
         stokes_u_cube=rmsynth_field_options.stokes_u_cube,
         rmsynth_options=rmsynth_options,
         stokes_i_cube=rmsynth_field_options.stokes_i_cube,
+        stokes_i_error_cube=rmsynth_field_options.stokes_i_error_cube,
```
AlecThomson: @claude - we won't use this feature here - not required

flint/options.py
```diff
@@ -259,14 +259,16 @@ class RMSynthOptions(BaseOptions):
     """Briggs robust parameter, required if weight_type is 'briggs'"""
     nufft_nthreads: int = 1
     """finufft OpenMP threads per dask chunk"""
+    reuse_rmsf: bool = True
+    """Compute one RMSF per chunk and reuse it for every pixel when they all share the same channel flagging, instead of running the NUFFT once per pixel for the same answer. rm-lite checks the condition per chunk, so a chunk with blanked pixels mixed in still gets per-pixel RMSFs, and this only ever declines a saving rather than asserting one. Worth ~20x on the RMSF stage for a typical cube, where flagging is per-channel and so shared by every pixel"""
```
AlecThomson: @claude - this docstr is far too verbose

flint/options.py
```diff
@@ -259,14 +259,16 @@ class RMSynthOptions(BaseOptions):
...
     target_chunk_mb: float = 256
     """Target per-chunk memory footprint, in MB, when reading the Q/U cubes"""
     fit_order: int = 2
     """Stokes I fractional-polarisation fit order; negative iterates orders and picks the best by AIC"""
     fit_function: Literal["log", "linear"] = "log"
     """Stokes I fit function: 'log' is a power law, 'linear' is a polynomial"""
     stokes_i_snr_cut: float | None = 5.0
-    """Below this frequency-averaged Stokes I SNR a pixel falls back to a flat model. None fits every pixel"""
+    """Below this frequency-averaged Stokes I SNR a pixel falls back to a flat model (no spectral correction, not blanked). None fits every pixel. Needs a Stokes I noise to compare against, i.e. ``estimate_stokes_i_noise`` or ``RMSynthFieldOptions.stokes_i_error_cube``; with neither, rm-lite's SNR is infinite for every pixel and this cut silently does nothing. That is the difference between fitting the handful of real sources and running a bounded ``curve_fit`` on every noise pixel in the cube (~25 ms each), so ``flint.rmsynth.run_rmsynth_3d`` warns when the combination is a no-op"""
```
AlecThomson: @claude - this docstr also is far too verbose. Revert it

flint/options.py
```diff
@@ -275,8 +277,10 @@ class RMSynthOptions(BaseOptions):
     """Also compute a debiased (via rm_lite's debias_fdf) mom0/mom1/mom2 set per requested FDF"""
     debias_filter_size: int = 5
     """Median filter size (pixels) used by mom0 debiasing"""
-    estimate_stokes_i_noise: bool = False
-    """Derive a per-channel Stokes I error from the Stokes I cube when fitting the fractional-polarisation model"""
+    moment_threshold_snr: float = 5.0
+    """SNR cut (times the theoretical FDF noise) applied to FDF amplitudes before flint computes any Faraday moment map, whichever FDF they come from. Without a cut, mom0 integrates |FDF| noise over every Faraday depth and so has a large positive floor everywhere (mom1/mom2 become meaningless with it), which is why this applies to the dirty moments too and not just the cleaned ones. Distinct from the identically named ``RMCleanOptions`` field, which only reaches rm-lite's own moment maps -- flint computes its own, see ``flint.rmsynth.write_rm_products``"""
```
AlecThomson: @claude - this docstr is far too verbose

flint/options.py
```diff
@@ -275,8 +277,10 @@ class RMSynthOptions(BaseOptions):
...
+    estimate_stokes_i_noise: bool = True
+    """Derive a per-channel Stokes I error from the Stokes I cube (a robust MAD per channel plane) when fitting the fractional-polarisation model. On by default because ``stokes_i_snr_cut`` is inert without a noise estimate, see there. Ignored when ``RMSynthFieldOptions.stokes_i_error_cube`` supplies a per-pixel error instead"""
```
AlecThomson: @claude - this docstr is far too verbose

flint/options.py
```diff
@@ -293,9 +297,21 @@ class RMCleanOptions(BaseOptions):
     gain: float = 0.1
     """CLEAN loop gain"""
     moment_threshold_snr: float = 5.0
-    """SNR cut (times the theoretical FDF noise) applied before computing Faraday moment maps"""
+    """SNR cut (times the theoretical FDF noise) handed to rm-lite for the moment maps it computes inside RM-CLEAN. Those maps are not what flint writes: ``flint.rmsynth.write_rm_products`` builds its own from the requested FDFs so they stay lazy, and cuts them at ``RMSynthOptions.moment_threshold_snr``"""
```
AlecThomson: @claude - this docstr is far too verbose

flint/options.py
```diff
@@ -293,9 +297,21 @@ class RMCleanOptions(BaseOptions):
...
     multiscale: bool = False
-    """Use multiscale RM-CLEAN, which can recover Faraday-thick structure"""
+    """Use multiscale RM-CLEAN, which can recover Faraday-thick structure. Experimental in rm-lite, and far slower than single-scale: it fits a Gaussian to every scale kernel of every pixel it cleans. Note rm-lite 2026.8.1 raises on a fully blanked (all-NaN) spectrum here, which a mosaic edge is full of"""
```
AlecThomson: @claude - this docstr is far too verbose

flint/options.py
```diff
@@ -351,6 +367,8 @@ class RMSynthFieldOptions(BaseOptions):
     """Path to the Stokes U FITS cube. Computed by the racs-all flow, so required only when running this pipeline standalone"""
     stokes_i_cube: Path | None = None
     """Path to a Stokes I FITS cube, used to fit a per-pixel fractional-polarisation correction. Defaults to None."""
+    stokes_i_error_cube: Path | None = None
+    """Path to a Stokes I error FITS cube matching ``stokes_i_cube``, used to weight the per-pixel fit and to set the per-pixel SNR the ``RMSynthOptions.stokes_i_snr_cut`` compares against. Takes precedence over ``RMSynthOptions.estimate_stokes_i_noise``, whose per-channel estimate is the fallback when this is None"""
```
AlecThomson: @claude - this docstr is far too verbose

flint/rmsynth.py
```diff
@@ -42,6 +44,16 @@
...
+# `reuse_rmsf` landed in rm-lite after 2026.8.1, the release flint currently
+# pins, so pass it only when the installed rm-lite actually accepts it. Either
+# way the answer is the same: without it rm-lite computes the RMSF per pixel,
+# which is what `reuse_rmsf=False` asks for anyway -- just slower. The saving
+# appears on its own once the pin moves to a release that has it, and this shim
+# (and the branch in `run_rmsynth_3d`) can go at that point.
+_RM_LITE_SUPPORTS_REUSE_RMSF = (
+    "reuse_rmsf" in inspect.signature(rmsynth_3d_from_fits).parameters
+)
+
```
AlecThomson: @claude Not required - we'll settle both at the same time

flint/rmsynth.py
```diff
@@ -75,11 +87,39 @@ def _check_cubes_memmappable(*cubes: Path | None) -> None:
...
+    """Warn when ``stokes_i_snr_cut`` is set but nothing gives it a noise to cut on.
+
+    rm-lite's frequency-averaged Stokes I SNR is ``mean(I) * sqrt(n) / rms(error)``,
+    and it deliberately returns infinity when the error spectrum is all-zero or
+    non-finite so that an SNR cut degrades to a no-op rather than rejecting
+    everything. With no error cube and no per-channel estimate the error *is*
+    all-zero, so every pixel scores infinity, passes the cut, and gets a full
+    bounded ``curve_fit`` -- around 25 ms on a pure-noise spectrum, which is
+    almost every pixel in a field cube. Loud rather than silent because the
+    result is correct but a thousand times slower than it needs to be.
```
AlecThomson: @claude Far too verbose

flint/rmsynth.py
```diff
@@ -88,15 +128,34 @@ def run_rmsynth_3d(
         stokes_u_cube (Path): Path to the Stokes U FITS cube
         rmsynth_options (RMSynthOptions): Options controlling the synthesis
         stokes_i_cube (Path | None, optional): Path to a Stokes I FITS cube, used to fit a per-pixel fractional-polarisation correction. FDF stays in Q/U flux if not given. Defaults to None.
+        stokes_i_error_cube (Path | None, optional): Path to a Stokes I error cube matching ``stokes_i_cube``, weighting the per-pixel fit and giving the SNR cut a per-pixel noise. Defaults to None, i.e. the per-channel estimate from ``RMSynthOptions.estimate_stokes_i_noise``.
```
AlecThomson: @claude drop this option

flint/rmsynth.py
```diff
@@ -88,15 +128,34 @@ def run_rmsynth_3d(
...
+    )
+
+    if stokes_i_cube is not None:
+        _warn_if_snr_cut_inert(rmsynth_options, stokes_i_error_cube)
+
+    reuse_rmsf_kwargs = (
+        {"reuse_rmsf": rmsynth_options.reuse_rmsf}
+        if _RM_LITE_SUPPORTS_REUSE_RMSF
+        else {}
+    )
```
AlecThomson: @claude drop

flint/rmsynth.py
```diff
@@ -133,6 +193,22 @@ def run_rmclean_3d(
...
+            else None,
+            "multiscale_n_scales": rmclean_options.multiscale_n_scales,
+            "multiscale_kernel": rmclean_options.multiscale_kernel,
+            "multiscale_max_iter_sub_minor": rmclean_options.multiscale_max_iter_sub_minor,
+            "multiscale_sub_minor_fraction": rmclean_options.multiscale_sub_minor_fraction,
+            "multiscale_selection_margin": rmclean_options.multiscale_selection_margin,
+        }
+        if rmclean_options.multiscale
+        else {}
+    )
```
AlecThomson: @claude - this seems extremely clunky

flint/rmsynth.py
```diff
@@ -303,41 +378,82 @@ def write_rm_products(
...
+        # model cubes ran RM-CLEAN twice per chunk, three cubes three times, and
+        # even one cube plus its moment maps twice -- although they all descend
+        # from a single `dask.delayed` call and are computed together below.
+        # Because the copy is made when the `Delayed` is built, `optimize_graph`
+        # at compute time cannot undo it; fusion has to be off both here and for
+        # the compute, since fusing there re-splits the branches again.
+        #
+        # Nothing is lost: these tasks are a whole spatial chunk of RM-CLEAN
+        # each, so saving a task boundary is worth far less than not running
+        # RM-CLEAN again.
```
AlecThomson: @claude far too verbose

CHANGELOG.md
AlecThomson, line 1: @claude we don't use this any more

flint/prefect/common/rmsynth.py
AlecThomson: Because it stopped being used. `write_rm_products` only ever read `moment_threshold_snr` off it, and that moved to `RMSynthOptions` so the cut could also apply to the dirty moment maps — which previously had no cut at all, leaving `mom0` with a noise floor of ~4.7 against an FDF noise of 0.11. With that gone, nothing in the function touched `rmclean_options`, so a required argument that callers had to pass and thread through the prefect task did nothing.

Happy to go the other way if you prefer: keep `rmclean_options` and use its `moment_threshold_snr` for every moment map, and drop the new field on `RMSynthOptions`. That is one option fewer overall, at the cost of the knob living under the `rmclean` strategy block even when you are only asking for dirty moments. Say which you'd rather and I'll switch it.

---

AlecThomson:
Review addressed in `aa023d2`. Docs and CHANGELOG reverted, docstrings trimmed, `stokes_i_error_cube` dropped throughout, the `multiscale_kwargs` dict replaced with plain keyword arguments, and the long comments cut. Diff is now 351 insertions across 6 files, down from 656 across 8.

**One consequence worth flagging, since it isn't visible in the diff:** dropping the rm-lite version probe means the branch now requires an rm-lite that has `reuse_rmsf`, and no release has it yet. So the rmsynth tests will fail against the pinned `rm-lite==2026.8.1` with `TypeError: rmsynth_3d_from_fits() got an unexpected keyword argument 'reuse_rmsf'` until the rm-lite change lands and the pin moves. I read "we'll settle both at the same time" as intending exactly that, but it does mean CI stays red in the interim rather than merely being unable to use the speed-up.

If you'd rather CI were green while the rm-lite side is in flight, the options are to drop `reuse_rmsf` from this PR and add it with the pin bump, or to put the probe back. Either is a small change — happy to do whichever you prefer.

Verified locally against the patched rm-lite: 22 tests passing across the RM-synthesis and flow suites, pre-commit clean.

The one thread I left open is the question about `rmclean_options` — I've replied there with why it went and an alternative, and didn't want to guess which way you'd want it.

---

flint/prefect/common/rmsynth.py
AlecThomson: @claude - then we are missing a whole bunch of options that should be being passed to rm-clean. Do a through check of this and reinstate rmcleanoptions

AlecThomson: Done in `1f7002e`. `rmclean_options` is back on `write_rm_products`, the prefect task and the flow, and its `moment_threshold_snr` is once again the single source for the moment cut — now also applied to the dirty maps, which is the part that was broken. The duplicate field I had added to `RMSynthOptions` is gone.

You were right that options were being dropped, and the audit pins down which. I compared both option classes against the rm-lite entry points in both directions:

**Was missing, now fixed.** Only the `multiscale` bool reached rm-lite; its six companion parameters (`scales`, `n_scales`, `kernel`, `max_iter_sub_minor`, `sub_minor_fraction`, `selection_margin`) had no route from a strategy file at all. This branch passes them.

**Also missing, and it was rm-lite's end.** `MultiscaleOptions.selection` (`"snr"` or `"hybrid"`, default `"hybrid"`) had no path through `run_rmclean` *or* `run_rmclean_from_synth`, so no 3D caller could set it. Exposed on the rm-lite branch, and added here as `RMCleanOptions.multiscale_selection`.

**Deliberately not rm-lite arguments.** `debias_moments` and `debias_filter_size` — flint builds the moment maps itself, so those are consumed in `write_rm_products`. Confirmed they are actually read, not dead.

**Left unset on purpose.** `log_level` is now the only rm-lite parameter neither class sets. The defaults (WARNING for synthesis, ERROR for CLEAN) are the right ones at cube scale — rm-lite logs per pixel below those, which floods. Happy to expose it if you want it reachable.

Every `RMCleanOptions` field now reaches rm-lite; the audit script is reproducible if you want it committed somewhere.

---

AlecThomson:
Correction to my earlier note, so the red CI isn't misread.

`1f7002e` adds a **second** dependency on unreleased rm-lite: `multiscale_selection` is passed to `run_rmclean_from_synth`, which the pinned `rm-lite==2026.8.1` also doesn't accept. That came out of the option audit rather than from review, and I should have flagged it as widening the version gap at the time rather than only in the commit body.

So the branch now needs an rm-lite with both `reuse_rmsf` (in `rmsynth_3d_from_fits`) and `multiscale_selection` (in `run_rmclean_from_synth`). Both are on the rm-lite branch; neither is in a release.

Confirmed the current red is one cause and nothing hiding behind it — the run on `aa023d2` was 13 failures, all `TypeError: rmsynth_3d_from_fits() got an unexpected keyword argument 'reuse_rmsf'`, with 444 passed and 8 skipped.

I'll leave it as-is rather than keep churning, since "settle both at the same time" reads as accepting this until the rm-lite side lands and the pin moves. If you'd rather have green CI in the meantime, either of these is a couple of minutes and I'm happy to do it on a word:

- move `reuse_rmsf` and `multiscale_selection` into a follow-up that lands with the pin bump, or
- gate both on the installed rm-lite signature, as `f597c92` did for `reuse_rmsf` before I dropped it.

---

flint/options.py
```diff
@@ -259,14 +259,16 @@ class RMSynthOptions(BaseOptions):
...
+    """Reuse one RMSF per chunk when every pixel shares the same channel flagging"""
     target_chunk_mb: float = 256
     """Target per-chunk memory footprint, in MB, when reading the Q/U cubes"""
     fit_order: int = 2
     """Stokes I fractional-polarisation fit order; negative iterates orders and picks the best by AIC"""
     fit_function: Literal["log", "linear"] = "log"
     """Stokes I fit function: 'log' is a power law, 'linear' is a polynomial"""
     stokes_i_snr_cut: float | None = 5.0
-    """Below this frequency-averaged Stokes I SNR a pixel falls back to a flat model. None fits every pixel"""
+    """Below this frequency-averaged Stokes I SNR a pixel falls back to a flat model. None fits every pixel. Needs ``estimate_stokes_i_noise``, else rm-lite scores every pixel as infinite SNR and the cut does nothing"""
```
AlecThomson: @claude "Needs ``estimate_stokes_i_noise``, else rm-lite scores every pixel as infinite SNR and the cut does nothing" this should be an error enforced in code, not just in the docs

flint/options.py
AlecThomson: Agreed, and done in `cbc9e7a`. `RMSynthOptions` now refuses `stokes_i_snr_cut` set alongside `estimate_stokes_i_noise=False`, and the docstring is back to the plain one-liner.

I put it on a pydantic `model_validator` rather than inside `run_rmsynth_3d`, for three reasons: it fires at option construction, which for a strategy file means at load rather than hours into a run; it can't be bypassed by any code path that builds the options differently; and `with_options` re-validates, so it survives the racs-all flow rebuilding them. The runtime warning it replaces is unreachable now and has gone.

`stokes_i_snr_cut=None` with no estimate is still allowed — that's a deliberate "fit every pixel", not the accident.

```
>>> RMSynthOptions(estimate_stokes_i_noise=False)
ValidationError: stokes_i_snr_cut=5.0 requires estimate_stokes_i_noise=True. Without a
Stokes I noise estimate rm-lite scores every pixel as infinite SNR, so the cut passes
all of them. Set estimate_stokes_i_noise=True, or stokes_i_snr_cut=None to fit every
pixel deliberately.
```

Tests cover the constructor, `with_options`, both allowed combinations, and the rule firing on options loaded from a strategy file.

---

flint/options.py
AlecThomson: Done in `cbc9e7a`, and present at `1ce42f3`: `RMSynthOptions` carries a pydantic `model_validator` that raises when `stokes_i_snr_cut` is set alongside `estimate_stokes_i_noise=False`, so the docstring note is now enforced.

A validator rather than a runtime check because it fires at option construction — for a strategy file that means at load, not hours in — and `with_options` re-validates, so it survives the racs-all flow rebuilding the options. `stokes_i_snr_cut=None` with no estimate is still allowed, as a deliberate fit-every-pixel.

I answered this on the neighbouring `options.py` thread at the time rather than here, which left this one open. Resolving it now.

---

flint/options.py
```diff
@@ -275,8 +298,8 @@ class RMSynthOptions(BaseOptions):
     """Also compute a debiased (via rm_lite's debias_fdf) mom0/mom1/mom2 set per requested FDF"""
     debias_filter_size: int = 5
     """Median filter size (pixels) used by mom0 debiasing"""
-    estimate_stokes_i_noise: bool = False
-    """Derive a per-channel Stokes I error from the Stokes I cube when fitting the fractional-polarisation model"""
+    estimate_stokes_i_noise: bool = True
+    """Derive a per-channel Stokes I error from the Stokes I cube when fitting the fractional-polarisation model. On by default because ``stokes_i_snr_cut`` is inert without it"""
```
AlecThomson: @claude - revert to original docstr

flint/options.py
```diff
@@ -293,9 +316,21 @@ class RMCleanOptions(BaseOptions):
...
+    # forbids extra fields, a strategy that sets `multiscale` under `rmclean` is
+    # refused at load rather than silently ignored, which is the point.
+    #
+    # To support it later: add `multiscale: bool = False` here and forward it in
+    # `flint.rmsynth.run_rmclean_3d`. rm-lite takes seven more parameters to tune
+    # it -- scales, n_scales, kernel, max_iter_sub_minor, sub_minor_fraction,
+    # selection and selection_margin -- each of which would want a
+    # `multiscale_`-prefixed field. `scales` wants a `list[float] | None`
+    # converted to an array at the call, since rm-lite takes an ndarray and None
+    # auto-selects.
```
AlecThomson: @claude - this comment is not needed. Delete.

flint/rmsynth.py
```diff
@@ -140,7 +148,9 @@ def run_rmclean_3d(
         max_iter=rmclean_options.max_iter,
         gain=rmclean_options.gain,
         moment_threshold_snr=rmclean_options.moment_threshold_snr,
-        multiscale=rmclean_options.multiscale,
+        # Multiscale RM-CLEAN is not supported for now, so nothing is passed for
+        # it and rm-lite's own `multiscale=False` applies; see `RMCleanOptions`
+        # for what enabling it here would take.
```
AlecThomson: @claude - comments not needed

flint/rmsynth.py
```diff
@@ -201,27 +211,75 @@ def _write(moment_set: FaradayMoments, suffix: str) -> list[Path]:
     "stokes_i_alpha_error": "stokesi.alpha_error",
     "stokes_i_model_order": "stokesi.model_order",
 }
+_STOKES_I_COEFF_SUFFIX = "stokesi.coeff"
```
AlecThomson: @claude - does this constant need to be global? If not, put as local to function

flint/rmsynth.py
```diff
@@ -201,27 +211,75 @@ def _write(moment_set: FaradayMoments, suffix: str) -> list[Path]:
...
+        )
+    if fit_function is not None:
+        header["FITFUNC"] = (fit_function, "Stokes I fit function")
+    if coeff is not None:
+        index, name = coeff
+        header["SICOEFF"] = (name, "Stokes I model term in this map")
+        header["SICOEFFI"] = (index, "Index of this term in the fitted popt")
+        header.add_comment("0 means the AIC dropped this term: it contributes nothing.")
+        header.add_comment("NaN means the pixel was never fitted (below the SNR cut).")
+    return header
```
AlecThomson, line 239: @claude - is it possible that this could drift from rm-lite? Or, are we safe to keep this here. If the former, we should import what we can from rm-lite

flint/rmsynth.py
```diff
@@ -232,6 +290,87 @@ def write_stokes_i_fit_maps_to_fits(
...
+    coeff_names: tuple[str, ...],
+    reference_header: fits.Header,
+    output_prefix: Path,
+    ref_freq_hz: float | None = None,
+    fit_function: str | None = None,
+    coeff_error_cube: np.ndarray | None = None,
+) -> list[Path]:
+    """Write the fitted Stokes I model terms, one FITS per named term rather than
+    one anonymous (n_coeff, ny, nx) cube.
+
```
AlecThomson, line 298: @claude 
```suggestion
    """Write the fitted Stokes I model terms.
```

flint/rmsynth.py
```diff
@@ -276,7 +415,7 @@ def write_rm_products(
         clean_results (RMClean3DResults | None): Results from ``run_rmclean_3d``, or None if 'clean'/'model' were not requested
         stokes_q_cube (Path): Path to the Stokes Q FITS cube (its header is reused for output WCS)
         rmsynth_options (RMSynthOptions): Options controlling RM-synthesis
-        rmclean_options (RMCleanOptions): Options controlling RM-CLEAN
+        rmclean_options (RMCleanOptions): Options controlling RM-CLEAN, whose ``moment_threshold_snr`` cuts every Faraday moment map
```
AlecThomson:
```suggestion
        rmclean_options (RMCleanOptions): Options controlling RM-CLEAN
```

flint/rmsynth.py
```diff
@@ -232,6 +290,87 @@ def write_stokes_i_fit_maps_to_fits(
...
+            np.asarray(coeff_cube[index], dtype=np.float32),
+            header,
+            overwrite=True,
+        )
+        output_paths.append(output_path)
+
+        if coeff_error_cube is None:
+            continue
+        # Marginal, i.e. sqrt(diag(pcov)): it ignores the strong correlations
+        # between the terms, so it is not the error on the model itself.
```
AlecThomson, line 352: @claude - this seems concerning. But is the issue upstream, or can we fix here?

AlecThomson (commented):
@claude - check for unnecessary global constants. You have a bad habit of creating them. They can and should be placed inside functions as local variables if they're only used in one place.

flint/rmsynth.py
AlecThomson: Safe to keep here, and there is nothing worth importing.

Every *value* in this header comes from rm-lite at runtime, not from a copy in flint. The term names are `synth_results.stokes_i_coeff_names`, the reference frequency is `stokes_i_ref_freq_hz`, and `fit_function` is flint's own option that rm-lite was called with. Flint never assumes a plane ordering either: it zips the names rm-lite returned against the cube's planes, and refuses with a clear message if the two counts disagree, so an rm-lite change to the number or order of terms surfaces as an error rather than as mislabelled maps.

rm-lite does export `coefficient_names(n_coeff, fit_function)` from `rm_lite.utils.fitting`, which is what it uses internally to build those names. Importing it would not add anything: it would recompute names flint already has on the results object, and if it ever disagreed with the cube it produced, the discrepancy would be rm-lite's, not something flint could resolve.

What *is* prose duplicated from rm-lite is the two `add_comment` lines describing its convention — zero means the AIC dropped a term, NaN means the pixel was never fitted. rm-lite exposes no constant for those, so there is nothing to import; a change of convention there would leave the wording stale. `test_stokes_i_model_rebuilds_from_the_written_term_maps` is the guard: it rebuilds the model from the written maps and their headers, checks it against rm-lite's own model cube, and asserts the AIC-dropped terms are exactly `0.0`. If rm-lite switched to NaN for a dropped term, that test fails. `test_rmsynth_writes_a_named_map_per_stokes_i_model_term` pins the names themselves for both fit functions, so a rename fails there too.

---

flint/rmsynth.py
AlecThomson: Neither, really — it is a limitation being labelled rather than a fault, and the correlation-aware alternative already exists upstream and is already exposed here.

`sqrt(diag(pcov))` *is* the marginal error by definition: the uncertainty on each term with the others free. It cannot be made into the error on the model, because that needs the off-diagonal covariance, and rm-lite returns only the diagonal (`stokes_i_coeff_error_cube`). So there is nothing to fix in flint, and the header comment exists so nobody quadrature-sums these to error-propagate `I(nu)` — for a curved power law the flux/alpha/beta terms are strongly anti-correlated, and doing that overestimates badly.

If you want a model error that does account for the correlations, rm-lite's route is `compute_model_error=True`, which Monte-Carlos the fit (`n_error_samples` resamples per pixel) and returns a model-error cube plus an alpha error. That is `RMSynthOptions.compute_model_error`, already reachable from a strategy; flint writes the alpha error as `stokesi.alpha_error` when it is on. It is off by default because it costs a thousand refits per fitted pixel.

The only thing that would need rm-lite is per-term errors that keep the correlations, which would mean returning `pcov` itself — an `(n_coeff, n_coeff, ny, nx)` array. Happy to raise that upstream if you want it, but the Monte-Carlo path covers the usual question.

---

flint/rmsynth.py
AlecThomson: No, and neither did its neighbour. Both are gone in `8137504`.

Auditing the three module-level constants in `flint/rmsynth.py`, per your review note:

- `_STOKES_I_COEFF_SUFFIX` — used only by `write_stokes_i_coeff_maps_to_fits`, now a local `suffix` there.
- `_STOKES_I_MAP_SUFFIXES` — used only by `write_stokes_i_fit_maps_to_fits`, now a local `suffixes` there. Its docstring named the constant, so that now lists the keys instead.
- `_MOMENT_NAMES` — read by both `write_moment_maps_to_fits` and `write_rm_products`, which have to agree on the order, so this one stays.

---

flint/options.py
AlecThomson: Done in `8137504`, back to the one-liner it was first added with:

```
"""Derive a per-channel Stokes I error from the Stokes I cube when fitting the fractional-polarisation model"""
```

Flagging the choice in case you meant the other one: this line had two earlier forms. The longer one also mentioned the MAD-per-channel-plane method and said it was ignored when `RMSynthFieldOptions.stokes_i_error_cube` supplied a per-pixel error — that option no longer exists, so restoring that text would have documented something gone. Say the word if you wanted the method mentioned and I will add just that clause.

---

PR #405, opened 2026-08-29
Fix RM-synthesis on the linmos weight cubes

AlecThomson:
Fixes the CI failure on `racs-all-pol`, plus two more breakages on the weight-cube path that only surface in a real run.

Pinned to **rm-lite 2026.8.5**, which carries the per-pixel weights refactor ([#70](https://github.com/AlecThomson/rm-lite/pull/70), [#71](https://github.com/AlecThomson/rm-lite/pull/71)) this branch is written against.

## 1. CI: `test_process_rmsynth_with_stokes_i_on_dask_cluster`

```
ValueError: snr_cut=5.0 needs a Stokes I error to measure SNR against,
and none was given (or it is all zero).
```

rm-lite 2026.8.4 added `check_snr_cut_has_error`. `8911540` dropped `RMSynthOptions.estimate_stokes_i_noise` on the assumption that the linmos Stokes I weight cube would always supply the error — but that test, and any standalone `process_rmsynth`, has a Stokes I cube and no weights.

Restored the option (default `True`). rm-lite's own precedence (`if stokes_i_error_file is not None … elif estimate_stokes_i_noise`) means a supplied weight cube still wins; the estimate only fills the gap. The validator that used to pair the option with the cut stays removed — rm-lite enforces that itself now.

## 2. The Q/U weight cubes

Wiring the linmos weights in broke RM-CLEAN on 2026.8.4: the theoretical FDF noise came back a lazy dask scalar, and `rmclean.py` formatted it with `:0.3g` → `TypeError: unsupported format string passed to Array.__format__`. A hard crash on any run with weight cubes.

2026.8.5 fixes that and goes further: **per-pixel weights are now the intended path**. Error cubes yield a lazy `(n_freq, ny, nx)` weight array, each pixel gets its own theoretical noise map, and the noise is formatted through `format_scalar_or_map`. So flint passes the cubes as `stokes_q_error_file`/`stokes_u_error_file` with `noise_files_are_weight=True` — linmos writes `1/sigma**2` directly, so rm-lite must not invert and square them again.

An earlier revision of this PR collapsed the cubes to a per-channel spectrum in flint. That is removed: it worked around the two bugs above, and after the refactor it would discard exactly the spatial information rm-lite is built to use.

**Two behaviour notes:**

- A pixel linmos blanked has no weight of its own and now comes back **blank**, rather than borrowing the rest of the field's channel weighting. An improvement — those pixels carry no data — but the FDF is NaN outside `pb_cutoff` where it previously wasn't.
- rm-lite **auto-enables the per-pixel RMSF** once pixels weight the channels differently, which a frequency-dependent primary beam guarantees. That roughly doubles the FDF's footprint, so it's documented on `RMSynthOptions.per_pixel_rmsf` and pinned by a test rather than left to surprise someone at scale. There is no way to decline it while using the weight cubes.

## 3. `PolPipelineResult` early return

The `wsclean_container is None` path never got `weight_cubes={}`, which is required, so it raised a `ValidationError` instead of returning empty. Both early returns now go through one `_no_products()` helper so a new field can't break them again.

## Also

- `per_pixel_rmsf` was added to `RMSynthOptions` in `8911540` but never passed to rm-lite — a silently ignored setting. Wired up.
- `stokes_i_weight_cube` was missing from the gzip-memmap check.

## Testing

- `pytest -ra -n 4 --skip-slow` → **469 passed** against the published 2026.8.5 wheel. Three failures are environmental to the sandbox, not this change: two Vizier tests (proxy blocks `vizier.cds.unistra.fr`) and `test_copy_directory` (no `rsync` in the container).
- `pre-commit run --hook-stage manual --all-files` → all hooks pass, mypy included.
- `flint/rmsynth.py` is at 100% patch coverage.

The Q/U weight path had **no** test coverage before this PR — the existing tests only ever passed `stokes_i_weight_cube`, which is exactly why the RM-CLEAN crash got past CI. It now covers:

| | |
|---|---|
| per-pixel theoretical noise map | shape and positivity |
| auto per-pixel RMSF | `rmsf_cube` present, RM-CLEAN consumes it |
| blanked primary-beam edge | blank outside the cutoff, finite inside |
| channels linmos flagged everywhere | drop out without moving the recovered φ |
| weights → moment maps | the path racs-all actually runs |
| a lone weight cube | rm-lite refuses it |

### DIFF
--- a/flint/options.py
+++ b/flint/options.py
@@ -276,7 +276,7 @@ class RMSynthOptions(BaseOptions):
     debias_filter_size: int = 5
     """Median filter size (pixels) used by mom0 debiasing"""
     per_pixel_rmsf: bool = False
-    """Compute the RMSF for each pixel rather than one shared by the cube. Roughly doubles the FDF's size"""
+    """Compute the RMSF for each pixel rather than one shared by the cube. Roughly doubles the FDF's size. rm-lite turns this on itself when the weights make pixels disagree, as the linmos weight cubes do"""
     estimate_stokes_i_noise: bool = True
     """Derive the per-channel Stokes I error from the Stokes I cube when no Stokes I weight cube is given. A weight cube takes precedence"""
 
--- a/flint/options.py
+++ b/flint/options.py
@@ -276,14 +276,9 @@ class RMSynthOptions(BaseOptions):
     debias_filter_size: int = 5
     """Median filter size (pixels) used by mom0 debiasing"""
     per_pixel_rmsf: bool = False
-    """Compute the RMSF for each pixel in the cube. ~2x the FDF's size, and only
-    worth turning on for per-pixel blanking that the channel weights do not carry"""
+    """Compute the RMSF for each pixel rather than one shared by the cube. Roughly doubles the FDF's size"""
     estimate_stokes_i_noise: bool = True
-    """Derive a per-channel Stokes I error from the Stokes I cube itself when no
-    Stokes I weight cube is given. rm-lite refuses a ``stokes_i_snr_cut`` with
-    nothing to measure SNR against, and scoring every pixel as infinite SNR would
-    fit noise as if it were signal, so the estimate is the fallback that keeps the
-    cut meaningful. A supplied weight cube takes precedence over it"""
+    """Derive the per-channel Stokes I error from the Stokes I cube when no Stokes I weight cube is given. A weight cube takes precedence"""
 
 
 class RMCleanOptions(BaseOptions):
--- a/flint/options.py
+++ b/flint/options.py
@@ -276,7 +276,14 @@ class RMSynthOptions(BaseOptions):
     debias_filter_size: int = 5
     """Median filter size (pixels) used by mom0 debiasing"""
     per_pixel_rmsf: bool = False
-    """ Compute the RMSF for each pixel in the cube """
+    """Compute the RMSF for each pixel in the cube. ~2x the FDF's size, and only
+    worth turning on for per-pixel blanking that the channel weights do not carry"""
+    estimate_stokes_i_noise: bool = True
+    """Derive a per-channel Stokes I error from the Stokes I cube itself when no
+    Stokes I weight cube is given. rm-lite refuses a ``stokes_i_snr_cut`` with
+    nothing to measure SNR against, and scoring every pixel as infinite SNR would
+    fit noise as if it were signal, so the estimate is the fallback that keeps the
+    cut meaningful. A supplied weight cube takes precedence over it"""
 
 
 class RMCleanOptions(BaseOptions):
--- a/flint/rmsynth.py
+++ b/flint/rmsynth.py
@@ -6,10 +6,12 @@ from __future__ import annotations
 
 import logging
 import os
+import warnings
 from pathlib import Path
 from typing import Any, Literal
 
 import dask
+import dask.array as da
 import numpy as np
 import zarr
 from astropy.io import fits
@@ -32,6 +34,7 @@ from rm_lite.tools_3d.rmsynth import (  # noqa: E402
     RMSynth3DResults,
     rmsynth_3d_from_fits,
 )
+from rm_lite.utils.dask_io import read_fits_cube_channel_chunks  # noqa: E402
 from rm_lite.utils.synthesis import (  # noqa: E402
     FaradayMoments,
     calc_faraday_moments,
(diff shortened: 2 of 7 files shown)


flint/options.py
```diff
@@ -276,7 +276,14 @@ class RMSynthOptions(BaseOptions):
...
     per_pixel_rmsf: bool = False
-    """ Compute the RMSF for each pixel in the cube """
+    """Compute the RMSF for each pixel in the cube. ~2x the FDF's size, and only
+    worth turning on for per-pixel blanking that the channel weights do not carry"""
+    estimate_stokes_i_noise: bool = True
+    """Derive a per-channel Stokes I error from the Stokes I cube itself when no
+    Stokes I weight cube is given. rm-lite refuses a ``stokes_i_snr_cut`` with
+    nothing to measure SNR against, and scoring every pixel as infinite SNR would
+    fit noise as if it were signal, so the estimate is the fallback that keeps the
+    cut meaningful. A supplied weight cube takes precedence over it"""
```
AlecThomson: @claude - simplify. This is a paragrpah

flint/rmsynth.py
```diff
@@ -77,6 +80,88 @@ def _check_cubes_memmappable(*cubes: Path | None) -> None:
...
+            "will be dropped from the RM-synthesis"
+        )
+    if not weight_arr.any():
+        msg = (
+            f"{stokes_q_weight_cube} and {stokes_u_weight_cube} are zero or blank "
+            "in every channel, so every channel would be dropped from the "
+            "RM-synthesis. Check that linmos produced usable weights."
+        )
+        raise NotSupportedError(msg)
+    return weight_arr
```
AlecThomson: @claude - this should be done in rm-lite

flint/options.py
AlecThomson: Done in 13eb6c0 — both new fields are single lines now, matching every other field in `RMSynthOptions`:

```python
per_pixel_rmsf: bool = False
"""Compute the RMSF for each pixel rather than one shared by the cube. Roughly doubles the FDF's size"""
estimate_stokes_i_noise: bool = True
"""Derive the per-channel Stokes I error from the Stokes I cube when no Stokes I weight cube is given. A weight cube takes precedence"""
```

The rationale that was in the paragraph already lives at the call site in `flint/rmsynth.py`, where it explains the kwarg rather than the option, so nothing was lost.

---

flint/rmsynth.py
AlecThomson: Agreed — the collapse belongs upstream, not here. Leaving it in place for now rather than removing it in this PR, because rm-lite `main` is currently exactly `v2026.8.4` (`0135cb3`), so dropping the helper today would put the RM-CLEAN `TypeError` back and re-break the smoke test with nothing to bump to.

I've posted the full spec in the session you're driving rm-lite from. The short version, against `main`:

1. **`rmsynth.py:617 get_weight_arr_from_fits`** — the error-file branch returns a lazy array, while the `get_noise_from_fits` fallback computes eagerly. That lazy value reaches `rmclean.py:324`, `f"Theoretical FDF noise: {fdf_error_noise:0.3g}"` → `TypeError: unsupported format string passed to Array.__format__`. Hard crash on any RM-CLEAN run with weight cubes.
2. **`rmsynth.py:128 _compute_global_params`** — a 3D `weight_arr` reaches the per-channel math and the whole-cube noise reduction. `_collapse_shared_weight_arr` (`:161`) already does the right thing, it's just applied too late (only in `_shared_rmsf`, `:201`).
3. **`rmsynth.py:161`** — that collapse uses `.mean(axis=(1,2))`, which linmos zeros outside `pb_cutoff` dominate. Ignoring non-positive/non-finite pixels (and keeping an all-blank channel at `0`, not `nan`) is what this PR's helper does.

Once those land and there's a tag to pin, `per_channel_weights_from_linmos` and its four tests come out of `flint/rmsynth.py`, and `run_rmsynth_3d` goes back to passing `stokes_q_error_file`/`stokes_u_error_file` with `noise_files_are_weight=True`. Happy to do that as a follow-up here, or hold this PR until the release — your call.

---

AlecThomson: Done in df8ed48 — `per_channel_weights_from_linmos` is gone and the weight cubes go straight to `stokes_q_error_file`/`stokes_u_error_file` again, as the original wiring did.

You were right, and more so than I'd argued: after `#70`/`#71` the collapse isn't just redundant, it's the wrong thing to do. Error cubes now yield a lazy `(n_freq, ny, nx)` weight array, each pixel gets its own theoretical noise, and `run_rmclean_from_synth` formats it via `format_scalar_or_map` instead of a bare `:0.3g`. Both problems the helper existed for are gone, and collapsing would throw away exactly the spatial information rm-lite is now built to use.

One genuine behaviour change worth knowing about: a pixel linmos blanked now has no weight of its own and comes back **blank**, rather than borrowing the rest of the field's channel weighting. That's an improvement — those pixels carry no data — but it does mean the FDF is NaN outside the cutoff where it previously wasn't. The end-to-end test now asserts blank-outside / finite-inside rather than finite-everywhere.

Also: rm-lite auto-enables the per-pixel RMSF once pixels weight the channels differently, which a frequency-dependent primary beam guarantees. That's the expensive half of per-pixel weights, so it's documented on `RMSynthOptions.per_pixel_rmsf` and pinned by a test rather than left to surprise someone at scale.

**CI will be red until you release.** 7 tests need the unreleased rm-lite (`_clean_block` → `_rmclean_on_block`, per-pixel noise, `rmsf_cube`). I've left `rm-lite==2026.8.4` alone for you to bump. Against `main` (`aa9fdfc`) the full suite is green: 467 passed, pre-commit clean. Say the word if you'd rather I pin a git SHA temporarily to get CI green in the meantime.

---

AlecThomson: Green on `50c59b7` with `rm-lite==2026.8.5` pinned — Python 3.11/3.12/3.13, Format, Distribution build, both Docker jobs, and codecov/patch. Resolving; the collapse is gone and the weights go to rm-lite as you asked.

Three rm-lite notes from working against 2026.8.5, none blocking:

1. **Cosmetic.** `rmsynth.py:961`, `stokes_i_error = 1 / da.sqrt(stokes_i_error)` emits `RuntimeWarning: divide by zero encountered in divide` wherever a linmos Stokes I weight is `0` — i.e. everywhere outside `pb_cutoff`, once per chunk. The result is right (pixel excluded, α is NaN there and −0.7 inside), but it will bury the logs at RACS scale. `_summarise_weight` already guards the identical case with `np.errstate(divide="ignore", invalid="ignore")`.

2. **No override on the auto per-pixel RMSF.** `per_pixel_rmsf` is force-set `True` when `not pixels_proportional`, with no way to say "I know, use the shared one". Defensible, but linmos weights guarantee it triggers, so the only lever left is dropping the weight cubes entirely. Worth knowing for memory planning.

3. **FYI.** `_summarise_weight` costs one extra full read of the weight cubes to decide proportionality. Bounded and streaming, so memory is fine — just real I/O on RACS-sized cubes.

Both duplicated-variable messages in `get_weight_arr_from_fits` (`{stokes_q_error_file=}` twice, and the Q/U-file equivalent) are fixed in this release.

---

PR #406, opened 2026-08-30
Report RM-synthesis progress as each product lands

AlecThomson:
Follow-up to #405. The RM-synthesis stage currently runs silent — this makes it report progress as each product lands.

> **Status:** the full suite is still running locally as I open this for review. Verified so far: `tests/test_rmsynth.py` 32 passed on rm-lite 2026.8.6, `pre-commit run --all-files` clean including mypy. The `DaskTaskRunner` flow tests (`test_prefect_rmsynth_flow.py`) are the outstanding check — I'll confirm below when they land, and push a fix if anything fails.

## The problem

The whole stage is one `dask.compute` over a lazy graph. `run_rmsynth_3d` and `run_rmclean_3d` return before anything is computed, so a run goes from "graph built" straight to "everything finished" — hours of silence on a mosaic, with no way to tell a slow run from a stuck one.

Logging around those calls would not help, and would actively mislead: `"RM-synthesis complete"` after `run_rmsynth_3d` would be a lie, since nothing has run at that point. The progress has to come from the compute itself.

## The change

`write_rm_products` submits the batch to the distributed `Client` as futures and drains them with `as_completed`:

```
Computing RM products (7 products over 5x5 pixels and 201 Faraday depths, in 5 chunks; per-pixel RMSF on)
[ 1/7] zarr_cubes at 0.5s
[ 2/7] moment.dirty.mom0 at 0.5s
...
Computed 7 RM products in 0.6s
```

The header line reports what the run is about to chew through — including whether rm-lite turned on the per-pixel RMSF, which the linmos weights guarantee and which flint never asks for (see #405). Elapsed is cumulative rather than per-product, since the products share one graph and "how far into the run are we" is the answerable question — hence `at`, not `in`.

Log volume is bounded by **product count, not chunks** (4–14 lines per run), so it cannot flood at RACS scale.

The local schedulers keep the plain `dask.compute` and are logged start-to-end only — `as_completed` needs a real `Client`, and that path is the unit tests and standalone use rather than a pipeline run.

## Two things this must not break

**The shared graph.** Everything is batched into one submission so the synthesis/RM-CLEAN graph runs *once per chunk* however many products are requested. Submitting per product multiplies the runtime invisibly — nothing in the output changes, it just takes N times longer.

The existing `test_rmclean_runs_once_per_chunk_whatever_is_requested` **cannot see this path**: it patches `dask.compute` to force the threaded scheduler. So there is a new `test_rmclean_runs_once_per_chunk_on_a_distributed_client`.

That test counts through a **file**, not a closure. `distributed` serialises the task graph even when its workers are threads in this process, so a list captured in a closure gets copied to the worker and the appends never come back — my first two attempts measured zero and looked like a pass.

I confirmed the guard actually bites by sabotaging the code to submit one compute per product: **20 RM-CLEAN calls against the expected 5**, caught and reported.

**Error propagation.** `dask.compute` raises the worker's exception directly; a future hands it back only when asked. A product that raised could otherwise go missing from the results and surface as a confusing `KeyError` downstream. `test_a_failed_product_is_raised_not_dropped` covers it.

## Also

Pins **rm-lite 2026.8.6**, which carries the fix for the divide-by-zero warning the linmos weight cubes provoked once per chunk everywhere outside `pb_cutoff`. I checked the no-API-change claim mechanically — public defs in `rmsynth.py`/`rmclean.py` and the result-tuple fields are identical to 2026.8.5 — and confirmed the warning is gone with behaviour unchanged (α = −0.7 inside the cutoff, NaN outside).

## Not included

Per-chunk granularity (`chunk 37/400 cleaned`) would need rm-lite: its block functions are silent and run on workers. Deliberately out of scope here.

No option is added — logging is unconditional at INFO, matching the rest of flint. Nothing about what is computed or written changes; the products and their contents are identical.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L2a7yxnqTQ1KBxanzoPMm8)_

### DIFF
--- a/flint/rmsynth.py
+++ b/flint/rmsynth.py
@@ -7,6 +7,8 @@ from __future__ import annotations
 import logging
 import os
 import time
+from collections.abc import Iterator
+from contextlib import contextmanager
 from pathlib import Path
 from typing import Any, Literal
 
@@ -15,7 +17,7 @@ import numpy as np
 import zarr
 from astropy.io import fits
 from astropy.wcs import WCS
-from dask.distributed import Client, as_completed
+from dask.distributed import Client, as_completed, get_worker, rejoin, secede
 
 # finufft's PyPI wheel vendors its own libomp.dylib (macOS) rather than
 # linking the environment's shared one; having both loaded aborts the
@@ -445,6 +447,37 @@ def _describe_rm_workload(
     )
 
 
+@contextmanager
+def _seceded_if_on_a_worker() -> Iterator[None]:
+    """Give this worker's compute slot back for the duration of a blocking wait.
+
+    ``task_write_rm_products`` is a prefect task, so under ``DaskTaskRunner`` it
+    runs *on a worker*, and ``prefect_dask.get_dask_client`` hands back a plain
+    ``Client`` rather than a ``worker_client`` -- it does not secede. Waiting on
+    futures from there holds the worker's thread while the tasks being waited on
+    need worker threads to run, and with one thread per worker that deadlocks the
+    cluster outright.
+
+    ``dask.compute(scheduler=client)`` never showed this because ``Client.get``
+    secedes for you when it is called inside a worker. Draining the futures by
+    hand is what makes it ours to do.
+
+    A no-op off a worker, which is the local-scheduler and standalone case.
+    """
+    try:
+        get_worker()
+    except ValueError:
+        # Not on a worker, so there is no slot to give back
+        yield
+        return
+
+    secede()
+    try:
+        yield
+    finally:
+        rejoin()
+
+
 def _compute_rm_products(
     compute_targets: dict[str, Any],
     fuse_config: dict[str, Any],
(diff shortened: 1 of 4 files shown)


AlecThomson:
Closing out the status banner in the description, which was written before the flow tests finished and is now stale in the record.

**They found a deadlock I introduced, fixed in `e9db832` before merge.**

`task_write_rm_products` is a prefect task, so under `DaskTaskRunner` it runs *on a worker*, and `prefect_dask.get_dask_client` hands back a plain `Client` rather than a `worker_client` — nothing secedes on our behalf. Draining futures from that thread while the tasks being waited on need worker threads is a deadlock, and at one thread per worker it is a certainty rather than a race. `dask.compute(scheduler=client)` never showed it because `Client.get` secedes for you when called inside a worker; draining by hand is what made it ours to do.

It surfaced as a hang, not a failure. py-spy on the live worker put its only compute thread here:

```
__next__ (distributed/client.py:5901)      <- as_completed, blocking
_compute_rm_products (flint/rmsynth.py:500)
task_write_rm_products (flint/prefect/common/rmsynth.py:81)
apply_function_simple (distributed/worker.py:3005)
```

CI was hanging on it too — the run on `a47ce5e` was cancelled after 26 minutes, so this was never local-only.

Fixed with `secede()`/`rejoin()` around the wait, a no-op off a worker. `test_computing_from_inside_a_worker_does_not_deadlock` is the whole reproduction in one worker with one thread, and fails by timeout in 90s with the secede removed.

**Final state on `e9db832`:** all 9 checks green — Python 3.11/3.12/3.13 at 5m25s/5m50s/5m06s, against a 4.3–5.8 min baseline and the 26-minute hang. Locally: flow tests 89.67s (~80s baseline), 56 passed across rmsynth + flow + racs-all, pre-commit clean including mypy.

One note for the record: the three lines codecov flags in `flint/rmsynth.py` are the off-worker branch of `_seceded_if_on_a_worker`. That path *is* exercised by every local-scheduler test, but coverage cannot attribute it — it is the body of a generator-based context manager. `codecov/patch` passed regardless.

---

PR #407, opened 2026-08-31
FDF peak statistics, RM-CLEAN niter map, and rm-lite 2026.8.7

AlecThomson:
Builds against [rm-lite v2026.8.7](https://github.com/AlecThomson/rm-lite/releases/tag/v2026.8.7) and wires up its new outputs.

## Peak statistics, per FDF

rm-lite 2026.8.7 returns FDF peak statistics on `RMClean3DResults` — but only for the clean FDF, and under its own threshold. The underlying `calc_faraday_peaks` is public and takes any complex FDF, exactly like `calc_faraday_moments`, which flint already uses to rederive its own moments.

So this adds `peak_products` as a third product list beside `cube_products` and `moment_products`, and every FDF can have all three kinds:

```
{prefix}.fdf.{dirty,clean,model}.peak_{pi,pi_debias,pi_error,rm,rm_error,pa,pa_error,pa0,pa0_error}.fits
```

Nine `(ny, nx)` maps per entry, so the list is **empty by default** — ~9 GB apiece at 16032².

Because dirty peaks come off the synthesis output, `needs_rmclean` takes `peak_products` the same way it takes the other two lists, and a dirty-only request correctly leaves RM-CLEAN out. That falls out of the list shape; it would have needed a special case as a boolean.

Each map carries a `BUNIT`: three of these are angles in degrees and two are Faraday depths, so a bare map is easy to misread.

## RM-CLEAN iteration count, unconditional

`{prefix}.fdf.clean.niter.fits`, written whenever RM-CLEAN runs rather than on request. It is one small integer map, and it is the one that says where CLEAN hit `max_iter` instead of converging on the threshold — a question asked *after* a run, when producing it on demand would mean repeating the stage. `int32` rather than rm-lite's `int64`, which still holds every value `max_iter` allows.

Both join the existing batched compute, so neither costs an extra pass over the graph — they show up as extra lines in the progress log from #406.

## `lam_sq_0_m2` in the strategy

Exposed on `RMSynthOptions` and passed through: `auto` derotates the cube to one reference λ², `per_pixel` gives each pixel its own (which also forces the per-pixel RMSF), or a float pins it.

## Where the options live

`peak_products` is on `RMCleanOptions`, not `RMSynthFieldOptions` — `RMSynthFieldOptions` is not in `MODE_OPTIONS_MAPPING`, so nothing on it is strategy-reachable, including the existing `cube_products`/`moment_products`. Putting it there is what makes it settable from a strategy file. The consequence is that the "no products requested" skip check has to run *after* the options are built rather than before.

Both new keys are in the example strategy with comments.

## Also

Pins `fitscube>=2.6` (2.5.1 → 2.6.0 in the lock).

## Verified, not just typechecked

- **Full 3×3 matrix** → 38 products (one zarr store + 9 moments + 27 peaks + the iteration map), and `peak_rm` recovers 50 rad/m² from all three FDFs: dirty 50.00, clean 49.99, model 49.73.
- **All nine peak maps come back finite**, including the five rm-lite documents as NaN without an `fdf_noise`. That was the open risk: since #405 flint's theoretical noise is an `(ny, nx)` map rather than a scalar, and it feeds them correctly.
- Iteration map is integer, capped by `max_iter`, and absent when RM-CLEAN did not run.
- `pytest -ra -n 4 --skip-slow` → **473 passed**. The only failures are the three environmental to the sandbox: two Vizier tests (proxy blocks `vizier.cds.unistra.fr`) and `test_copy_directory` (no `rsync` in the container).
- `pre-commit run --hook-stage manual --all-files` → clean, mypy included.

**Five pre-existing exact-count assertions changed** (four in `test_rmsynth.py`, one in `test_prefect_rmsynth_flow.py`). That is the intended behaviour of an unconditional iteration map — one more file per RM-CLEAN run — so the expectations moved, with comments saying why, rather than the code.

### DIFF
--- a/flint/data/tests/test_config_pol.yaml
+++ b/flint/data/tests/test_config_pol.yaml
@@ -118,6 +118,14 @@ polarisation:
 rmsynth:
   rmsynth:
     phi_max_radm2: 1000.0
+    # 'auto' derotates the whole cube to one reference lambda^2; 'per_pixel'
+    # gives each pixel its own, which also forces the per-pixel RMSF
+    lam_sq_0_m2: auto
   rmclean:
     auto_mask: 7
     auto_threshold: 1
+    # Peak statistics per FDF, like cube_products/moment_products: peak
+    # polarised intensity (raw and debiased), Faraday depth, polarisation angle
+    # and intrinsic angle, each with its error. Nine (ny, nx) maps per entry --
+    # ~9 GB apiece at 16032^2 -- so empty by default. 'dirty' needs no RM-CLEAN.
+    peak_products: []
--- a/flint/data/tests/test_config_pol.yaml
+++ b/flint/data/tests/test_config_pol.yaml
@@ -118,14 +118,10 @@ polarisation:
 rmsynth:
   rmsynth:
     phi_max_radm2: 1000.0
-    # 'auto' derotates the whole cube to one reference lambda^2; 'per_pixel'
-    # gives each pixel its own, which also forces the per-pixel RMSF
+    # 'auto', 'per_pixel', or a lambda^2 in m^2
     lam_sq_0_m2: auto
   rmclean:
     auto_mask: 7
     auto_threshold: 1
-    # Peak statistics per FDF, like cube_products/moment_products: peak
-    # polarised intensity (raw and debiased), Faraday depth, polarisation angle
-    # and intrinsic angle, each with its error. Nine (ny, nx) maps per entry --
-    # ~9 GB apiece at 16032^2 -- so empty by default. 'dirty' needs no RM-CLEAN.
+    # Any of: dirty, clean, model
     peak_products: []
--- a/flint/rmsynth.py
+++ b/flint/rmsynth.py
@@ -437,34 +437,6 @@ def _lazy_faraday_moments(
     )
 
 
-def _lazy_faraday_peaks(
-    fdf_cube: dask.array.Array,
-    synth_results: RMSynth3DResults,
-    threshold: float | np.ndarray | None,
-) -> FaradayPeaks:
-    """Build the lazy peak-statistic maps of an FDF cube.
-
-    The same shape of work as ``_lazy_faraday_moments``: ``calc_faraday_peaks``
-    reduces along the (never-chunked) Faraday-depth axis, so the result is nine
-    lazy (ny, nx) maps each spatial chunk contributes to independently, and the
-    cube itself never has to reach the calling worker.
-
-    rm-lite also returns peaks of its own on ``RMClean3DResults``, but only for
-    the clean FDF and only under its own threshold. Deriving them here instead
-    is what lets any requested FDF have them -- the dirty one included, which
-    needs no RM-CLEAN at all -- under the same cut flint gives its moments.
-    """
-    return calc_faraday_peaks(
-        fdf_cube,
-        phi_arr_radm2=synth_results.phi_arr_radm2,
-        fwhm_rmsf_radm2=synth_results.fwhm_rmsf_radm2,
-        fdf_error=synth_results.theoretical_noise.fdf_error_noise,
-        lam_sq_0_m2=synth_results.lam_sq_0_m2,
-        lambda_sq_arr_m2=synth_results.lambda_sq_arr_m2,
-        threshold=threshold,
-    )
-
-
 def _elapsed(start: float) -> str:
     """Wall time since ``start``, as a compact human-readable string."""
     seconds = time.time() - start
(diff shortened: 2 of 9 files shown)


flint/data/tests/test_config_pol.yaml
```diff
@@ -118,6 +118,14 @@ polarisation:
...
+    # 'auto' derotates the whole cube to one reference lambda^2; 'per_pixel'
+    # gives each pixel its own, which also forces the per-pixel RMSF
+    lam_sq_0_m2: auto
   rmclean:
     auto_mask: 7
     auto_threshold: 1
+    # Peak statistics per FDF, like cube_products/moment_products: peak
+    # polarised intensity (raw and debiased), Faraday depth, polarisation angle
+    # and intrinsic angle, each with its error. Nine (ny, nx) maps per entry --
+    # ~9 GB apiece at 16032^2 -- so empty by default. 'dirty' needs no RM-CLEAN.
```
AlecThomson: @claude - rewrite this comment to just document what values can go in here. This current comment is too long and isn't useful

flint/rmsynth.py
```diff
@@ -47,21 +49,40 @@
 FDFLabel = Literal["dirty", "clean", "model"]
 _MOMENT_NAMES = ("mom0", "mom1", "mom2")
 
+# The FDF peak statistics, as {FaradayPeaks field: (file suffix, BUNIT, comment)}.
+# Units matter more here than for the moments: three of these are angles in
+# degrees and two are Faraday depths, so a map with no BUNIT is easy to misread.
+_PEAK_MAPS = {
```
AlecThomson, line 55: @claude - does this need to be global? Same for the above. Does it need to be private?

flint/rmsynth.py
```diff
@@ -415,6 +437,34 @@ def _lazy_faraday_moments(
...
+    """
+    return calc_faraday_peaks(
+        fdf_cube,
+        phi_arr_radm2=synth_results.phi_arr_radm2,
+        fwhm_rmsf_radm2=synth_results.fwhm_rmsf_radm2,
+        fdf_error=synth_results.theoretical_noise.fdf_error_noise,
+        lam_sq_0_m2=synth_results.lam_sq_0_m2,
+        lambda_sq_arr_m2=synth_results.lambda_sq_arr_m2,
+        threshold=threshold,
+    )
```
AlecThomson: @claude what does this actually do?

AlecThomson: That is, it seems useless to me

AlecThomson: You're right — it forwarded every argument unchanged and had exactly one caller, so it was a docstring with a function around it. Inlined at the call site in `51febb9`.

The one part worth keeping was the *why*, so that survives as a comment there: rm-lite returns peaks of its own on `RMClean3DResults`, but only for the clean FDF and only under its own threshold, so deriving them in flint is what lets `dirty`/`clean`/`model` each have them under the same cut flint gives its moments.

I left `_lazy_faraday_moments` alone — it has two callers with different arguments and one conditional, so it isn't the same pure passthrough.

---

flint/data/tests/test_config_pol.yaml
AlecThomson: Trimmed in `51febb9`, to just:

```yaml
    # Any of: dirty, clean, model
    peak_products: []
```

I applied the same to the `lam_sq_0_m2` comment I'd added a few lines up, since it had the same problem — it's now `# 'auto', 'per_pixel', or a lambda^2 in m^2`. The longer explanation of each already lives on the option docstrings in `flint/options.py`.

---

flint/rmsynth.py
AlecThomson: **Global: yes, both are genuinely shared.**

- `_MOMENT_NAMES` — read in `write_moment_maps_to_fits`, and twice each in the batching and the writing halves of `write_rm_products`.
- `_PEAK_MAPS` — read in `write_peak_maps_to_fits` and in both halves of `write_rm_products`.

That sharing is the point rather than a convenience: the batching side keys `compute_targets` off these names and the writing side looks the results back up by the same keys, so a local copy in either function would let the two drift into a `KeyError`.

**Private: yes** — nothing outside `flint/rmsynth.py` reads either one (`FDFLabel` is the name that crosses the module boundary, and it's already public). Happy to drop the underscores if you'd rather they be discoverable as the list of available products, but I'd default to not advertising an API flint doesn't consume.

One thing your question did shake out: `_PEAK_MAPS`' key order *was* load-bearing, because `write_rm_products` rebuilt `FaradayPeaks` positionally from it. It happens to match rm-lite's field order today, but a reorder upstream would have silently mis-assigned all nine maps. It's keyword-constructed as of `51febb9`.

---

PR #409, opened 2026-09-01
Bring the RM-synthesis input cubes to a common resolution

AlecThomson:
RM-synthesis is only meaningful when every channel of every Stokes shares one restoring beam, but the polarisation stage convolves each channel of each Stokes to its own common beam, so the cubes it hands over rarely do.

`process_rmsynth` now checks the Stokes cubes before running the synthesis, which covers both the standalone pipeline and the racs-all meta flow (both go through this flow). The work is done by `racs_tools`, not re-implemented here.

### If the cubes already share a resolution, they are used as they are

`cubes_share_common_beam` reads the per-channel beams through `racs_tools.beamcon_3D.make_data` (which already handles the CASA beam table, beamlog and header-beam fallbacks) and compares them with `radio_beam`'s own `Beam` equality. Cubes with no beam information at all have no resolution to make common.

### Otherwise new cubes are made, without touching the inputs

`convolve_cubes_to_common_beam` hands the cubes to `smooth_fits_cube` in `total` mode — the one beam across all channels and all cubes, as opposed to `natural`'s beam per channel. It derives the beam, convolves, and returns the new `.conv.fits` cubes. The inputs are never overwritten: the convolved cubes go into the rm-synth `output_path` (alongside the inputs when there is none) and the resolved paths are what the synthesis reads. Convolution preserves the pixel grid, so the LINMOS weight cubes are left untouched.

Two wrinkles handled at the boundary:

- `smooth_fits_cube` sorts its inputs, so its outputs are paired back to each Stokes by name rather than by position.
- `total` mode copies the input header's `CASAMBM` flag into a cube that now has a single beam and no `BEAMS` table, so that stale flag is cleared.

### The new cubes are spiced alongside the cubes they came from

Left as it was, the racs-all flow would strand them: `process_rmsynth` wrote them into the rm-synth directory and returned only the RM products, so the spice stage never saw them — it trimmed the polarisation cubes and removed the originals, leaving a full-size, uncompressed copy of every Stokes cube on disk.

`process_rmsynth` now returns an `RMSynthPipelineResult` carrying the written products and any convolved cubes, matching how `PolPipelineResult` already reports upward, and racs-all appends those to the cubes it hands the spice stage. One set of island boxes and one weight set cover both, since convolution preserves the pixel grid. `convolved_cubes` is empty when the inputs already shared a resolution, so concatenating it can never repeat a path — a repeat would be spiced twice and its input unlinked twice. The skip-spice compression fallback covers them too.

### Island boxes are sized on the cubes being trimmed

`process_spice_compression` sized them on `Beam.from_fits_header(reference_image)` — the Stokes I MFS image. This predates the work above: the polarisation stage convolves each channel to its own common beam, so the coarsest channel is already broader than the MFS beam and the boxes have been undersized at the low-frequency end. They are now sized on `common_beam_from_cubes` over the cubes themselves, which also removes the edge where `reference_image` is a multi-beam cube carrying no `BMAJ` in its header.

### Optional beam cutoff

`RMSynthFieldOptions.beam_cutoff` is threaded through to `racs_tools`, which blanks any channel coarser than it and leaves it out of the common beam rather than dragging every channel out to that resolution. Unset by default, since the racs-all flow already applies `PolFieldOptions.beam_cutoff` when convolving each channel in the polarisation stage. Note that the two stages share the field name, so the combined racs-all CLI exposes one `--beam-cutoff` feeding both.

### Dependency bump

`RACS-tools>=4.3.2`, for two fixes this PR is the first to need:

- 4.3.0 cannot write a convolved cube under numpy 2 — `initfiles` does `int(spec_axis.wcs.crpix)` on a size-1 array, which numpy 2 rejects. Latent until now, since `convolve_cubes` had no callers and its test is commented out.
- 4.3.1 leaves its module-level `QueueListener` started (it only enqueued a sentinel, never stopped the listener), so the second `smooth_fits_cube` call in a process raised `RuntimeError: Listener already started` under python 3.13. That broke `flint_convol --mode convol --cubes`, which dry-runs for the common beam and then convolves in the one process. [4.3.2](https://github.com/AlecThomson/RACS-tools/releases/tag/v4.3.2) runs the listener under a `running_log_listener()` context manager, so it is stopped on every exit path.

### Tests

- `tests/test_convol.py`: the share/no-share cases across channels and across Stokes, cubes with no beams at all, `common_beam_from_cubes`, and an end-to-end `convolve_cubes_to_common_beam` asserting the inputs survive, the outputs land in the output directory, the shape is preserved and the results share one beam.
- `tests/test_prefect_rmsynth_flow.py`: the flow-level resolution step, both when it must convolve (including that each Stokes keeps its own data through the sort) and when it must not, plus the beam cutoff blanking a coarse channel instead of convolving to it.
- `tests/test_racs_all_pipeline.py`: `process_racs_all` with the stage subflows mocked, asserting the convolved cubes reach the spice stage and that nothing is repeated when no convolution happened.

https://claude.ai/code/session_01MzDY78jjEiVKdumjYTEjms


flint/convol.py
```diff
@@ -80,6 +81,93 @@ def check_if_cube_fits(fits_file: Path) -> bool:
     return len(squeeze_data.shape) == 3
 
 
+def _beams_from_cubes(cube_paths: list[Path]) -> list[Beam]:
```
AlecThomson: @claude - why a `list[Beam]` and not a `Beams`?

AlecThomson: No good reason — fixed in 0ddfd04. `_beams_from_cubes` now returns a `Beams`, which makes `common_beam_from_cubes` a one-liner off it, and `cubes_share_common_beam` needed no change since indexing and iterating a `Beams` both yield `Beam`.

---

flint/convol.py
```diff
@@ -80,6 +81,122 @@ def check_if_cube_fits(fits_file: Path) -> bool:
...
+            # The sentinel racs_tools sends on its way out has already asked the
+            # monitor to finish, so joining it leaves nothing in the queue. A
+            # dry run sends no sentinel, and only there is stop() needed -- it
+            # enqueues one that no monitor is left to read, which would kill the
+            # next listener the moment it starts.
+            thread.join(timeout=10)
+            if thread.is_alive():
+                log_listener.stop()
+            else:
+                log_listener._thread = None
```
AlecThomson: @claude why is this needed? Is this just for the tests?

AlecThomson:
Answering https://github.com/flint-crew/flint/pull/409#discussion_r3900950897 here — a pending review on the PR blocks review-thread replies, so this couldn't go on the thread.

> why is this needed? Is this just for the tests?

Not just the tests — it breaks a CLI path that exists today, independently of this PR.

`racs_tools.logging` holds one module-level `QueueListener`. `smooth_fits_cube` calls `log_listener.start()` on entry but only `enqueue_sentinel()` on exit, never `stop()`, so the finished thread stays recorded on the listener. Python 3.13 added a guard rejecting the next `start()`, so the *second* `smooth_fits_cube` call in a process raises. The dry-run path is worse: it returns before the sentinel, so its thread is still running.

`flint_convol --mode convol --cubes` does exactly that — `get_cube_common_beam` (a dry run) and then `convolve_cubes` per image, all in one process. Installing 3.13's guard on 3.11 and running that sequence with the wrapper removed:

```
STEP1 get_cube_common_beam ok
REPRODUCED: RuntimeError: Listener already started
```

and with it in place:

```
STEP1 get_cube_common_beam ok
STEP2 convolve_cubes q.fits ok
STEP2 convolve_cubes u.fits ok
```

A dask worker also outlives the task it ran, so a worker convolving twice hits the same thing. This PR just added the first caller that runs more than once inside a pytest process, which is how it surfaced.

`e1c1293` puts that in the docstring, since the code didn't say it. The test walks the CLI's own order (dry run, then convolve) with the guard installed, so 3.11 and 3.12 cover it rather than only 3.13 catching a regression.

On why it isn't simply `stop()`: that enqueues a second sentinel no monitor is left to read, and the next listener would read the stale one and exit the moment it started — silently ending the worker log relay, with a sentinel accumulating per call. So the finished thread is joined and cleared, and `stop()` is kept for the dry run, the only case that leaves the thread alive.

The real fix belongs upstream: `smooth_fits_cube` and `smooth_fits_files` should call `stop()`, or build a fresh listener per call. I can't open that issue — this session's GitHub access is scoped to `flint-crew/flint` — so happy for this to stand as a stopgap pointing at an upstream issue if you file one.

---

pyproject.toml
```diff
@@ -44,7 +44,7 @@ dependencies = [
     "prefect>=3.4.7",
     "prefect-dask>=0.3.7",
     "dask-jobqueue",
-    "RACS-tools>=4.2.1",
+    "RACS-tools>=4.3.2", # 4.3.0 cannot write a convolved cube under numpy 2; 4.3.1 leaves its log listener started, which python 3.13 rejects on the next call
```
AlecThomson:
```suggestion
    "RACS-tools>=4.3.2"
```

PR #412, opened 2026-09-01
Convolve the RM-synthesis cubes plane by plane, and never solve an empty beam set

AlecThomson:
Fixes the `ValueError('attempt to get argmax of an empty sequence')` raised from
`convolve_cubes_to_common_beams` in the rm-synthesis flow, and reworks the
convolution so it can actually use the `dask` cluster.

## The crash

`radio_beam`'s common-beam solver reduces a beam set to the beams it considers
finite (`major > 0 & minor > 0 & isfinite(pa)`) and takes an `argmax` over what
is left. `beamcon_3D`'s `total` mode masks blanked channels and channels beyond
the cutoff itself, then hands the remainder straight to `common_beam()` -- so a
set in which no channel carries a beam `radio_beam` can use reduces to an empty
sequence and raises from inside `astropy`, with nothing in the traceback saying
which cube or channel was at fault.

`flint` was handing it such a set. The polarisation stage convolves every
channel of every beam to one 2D common beam, so the cubes it produces normally
carry a single real beam plus a placeholder for each blanked channel -- exactly
zero (wsclean's convention for a plane with no fitted PSF) or the `float32` tiny
that `fitscube` writes into a BEAMS table, as CASA rejects a NaN there. The old
`cubes_share_common_beam` counted those placeholders as a beam mismatch, so it
convolved cubes that were already at one resolution, which is why the step ran
at all.

## Plane by plane, not cube by cube

`convolve_cubes_to_common_beam` handed whole cubes to `beamcon_3D.smooth_fits_cube`,
which walks one cube at a time on a single worker and pulls each cube into memory
to convolve it in 3D. `convolve_cubes_to_common_resolution` (in
`flint.prefect.common.rmsynth`) instead splits every cube into its per-channel
planes, convolves each plane as its own prefect task, and stacks the planes back
into a cube. Every channel of every Stokes is smoothed at once across the
cluster, and no task holds more than a single plane. This is what
`create_convolve_linmos_cubes` already does for the per-beam cubes.

- Each cube is split into its own `<stem>.planes/` directory, as a plane's name
  carries only the flint name fields and cubes agreeing on those would clobber
  each other's planes. The planes are removed once cubed.
- The output is `<input stem>.conv.cube.fits`, assembled with
  `bounding_box=False` and `create_blanks=False` so the untouched weight cubes
  still match the pixel and channel grid.
- `split_cube_into_planes` takes an `output_path`, defaulting to alongside the
  cube as before.
- The unused `task_split_cube_into_planes` in `imaging.py` is replaced by a
  wrapper of the real function, rather than leaving two prefect tasks of that
  name in sibling modules.

## Filtering the beams before solving

`usable_beam_mask` excludes placeholder beams, non-finite beams, and beams
beyond the cutoff -- a strictly stronger test than `radio_beam`'s, so what is
handed to `common_beam()` can never reduce to an empty sequence.

- `common_beam_from_cubes` returns `None` when no channel of any cube carries a
  real beam, and the rm-synthesis flow then uses the cubes as they are.
- `cubes_share_common_beam` compares only usable beams, so cubes the
  polarisation stage already brought to one beam are no longer reconvolved
  because some channels are blank. A channel with a real beam beyond the cutoff
  is still reported as *not* common, since blanking it is work only the
  convolution pass does.
- `process_spice_compression` raises a named `ValueError` on `None`; it would
  otherwise have hit the same `argmax` error when sizing its island boxes.
- `convolve_images` skips an image whose own beam is a placeholder rather than
  only one whose `BMAJ` is exactly zero, and a plane blanked by the cutoff is
  marked with a zero beam so the cube it is stacked into does not claim a PSF
  for a channel holding none.

## Tests

The cube fixtures in `tests/test_convol.py` and
`tests/test_prefect_rmsynth_flow.py` are now flint-named, which splitting
requires and the flow already required for `create_name_from_common_fields`.
Added coverage for the usable-beam mask, per-plane convolution (including the
cutoff and no-beam cases), placeholder beams in the share/common-beam tests, and
the all-placeholder cubes that used to crash.

163 tests across `test_convol`, `test_prefect_rmsynth_flow`,
`test_racs_all_pipeline`, `test_wsclean`, `test_prefect_racsall_flow`,
`test_prefect_helpers`, `test_spice` and `test_naming` pass; `ruff` is clean and
`mypy` reports no new errors.

## Worth noting

With placeholder beams no longer counting as a mismatch, the rm-synthesis stage
will usually now skip the convolution on real racs-all output and pass the
polarisation cubes straight through. That is the intended design, but it does
mean the new plane path will not see much use in production until the input
cubes genuinely differ in resolution.

https://claude.ai/code/session_01Bonk4muETYXyG3E6knCfDu

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bonk4muETYXyG3E6knCfDu)_

### DIFF
--- a/flint/convol.py
+++ b/flint/convol.py
@@ -16,8 +16,10 @@ import numpy as np
 from astropy.io import fits
 from astropy.wcs import FITSFixedWarning
 from racs_tools import beamcon_2D, beamcon_3D
+from racs_tools.convolve_uv import my_ceil, round_up
 from radio_beam import Beam, Beams
 from radio_beam.beam import NoBeamException
+from radio_beam.utils import BeamError
 
 from flint.logging import logger
 
@@ -217,7 +219,26 @@ def common_beam_from_cubes(
         return None
 
     logger.info(f"Deriving a common beam from {usable.sum()} of {len(usable)} beams")
-    return beams[usable].common_beam()
+    usable_beams = beams[usable]
+    try:
+        common_beam = usable_beams.common_beam()
+    except BeamError:
+        logger.warning(
+            "Could not find a common beam with the default tolerance, trying again"
+        )
+        common_beam = usable_beams.common_beam(tolerance=1e-5)
+
+    # The minimum enclosing ellipse radio_beam solves for sits right against the
+    # beams it encloses, so a channel whose own beam it barely covers can fail to
+    # deconvolve. Rounding up gives every channel the headroom to reach it, which
+    # is what racs_tools does with the common beams it derives itself.
+    return Beam(
+        major=my_ceil(common_beam.major.to(u.arcsecond).value, precision=1)
+        * u.arcsecond,
+        minor=my_ceil(common_beam.minor.to(u.arcsecond).value, precision=1)
+        * u.arcsecond,
+        pa=round_up(common_beam.pa.to(u.degree).value, decimals=2) * u.degree,
+    )
 
 
 def convolve_plane_to_beam(
--- a/flint/convol.py
+++ b/flint/convol.py
@@ -23,6 +23,13 @@ from flint.logging import logger
 
 warnings.simplefilter("ignore", FITSFixedWarning)
 
+# A restoring beam smaller than this is not a real PSF. Blank channels are
+# marked either with an exactly-zero beam (wsclean's convention for a plane
+# with no fitted beam) or with ``np.finfo(np.float32).tiny`` (the sentinel
+# ``fitscube`` writes into a BEAMS table, as CASA rejects a NaN there). Both
+# sit many orders of magnitude below any beam an interferometer can produce.
+MIN_USABLE_BEAM_ARCSEC = 1e-6
+
 
 class BeamShape(NamedTuple):
     """A simple container to represent a fitted 2D gaussian,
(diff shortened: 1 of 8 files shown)


flint/prefect/common/rmsynth.py
```diff
@@ -22,7 +29,103 @@
     write_rm_products,
 )
 
-task_convolve_cubes_to_common_beam = task(convolve_cubes_to_common_beam)
+task_convolve_plane_to_beam = task(convolve_plane_to_beam)
+
+CONVOL_SUFFIX = "conv"
+"""Marker added to the name of a cube brought to a common resolution"""
+
```
AlecThomson: @claude why is this global?

flint/prefect/common/rmsynth.py
```diff
@@ -22,7 +29,103 @@
...
+    create_blanks=False,
+    # Convolution has already put the pixel values on a new scale, so leaving
+    # exact zeros be avoids reinterpreting them a second time
+    invalidate_zeros=False,
+    # rm-synth and the spice stage both read these cubes plane by plane, and
+    # astropy cannot memmap a gzip file
+    compress=False,
+    remove_original_images=True,
+)
+"""How a cube is reassembled from its convolved planes"""
```
AlecThomson: @claude why is this global?

flint/convol.py
```diff
@@ -16,13 +16,22 @@
...
 
 warnings.simplefilter("ignore", FITSFixedWarning)
 
+# A restoring beam smaller than this is not a real PSF. Blank channels are
+# marked either with an exactly-zero beam (wsclean's convention for a plane
+# with no fitted beam) or with ``np.finfo(np.float32).tiny`` (the sentinel
+# ``fitscube`` writes into a BEAMS table, as CASA rejects a NaN there). Both
+# sit many orders of magnitude below any beam an interferometer can produce.
+MIN_USABLE_BEAM_ARCSEC = 1e-6
+
```
AlecThomson: @claude why is this global?
Also, why not use np.finfo(np.float32).tiny?

flint/prefect/common/rmsynth.py
AlecThomson: No good reason — it was used once, inside `convolve_cubes_to_common_resolution`. It is now a `convol_suffix: str = "conv"` parameter of that function, matching `convolve_images` and `convolve_plane_to_beam`, which it is handed straight to. Fixed in 49f0db9.

---

flint/prefect/common/rmsynth.py
AlecThomson: Same — used once, so it is now a local inside `convolve_cubes_to_common_resolution`, comments and all. Fixed in 49f0db9.

Worth flagging what it is doing there, since the settings are not the usual ones: `bounding_box=False` and `create_blanks=False` because the weight cubes are not convolved, so these cubes have to come back on the same pixel and channel grid as the inputs; `invalidate_zeros=False` because convolution has already rescaled the pixels; `compress=False` because rm-synth and spice both read these plane by plane and astropy cannot memmap a gzip file.

---

flint/convol.py
AlecThomson: Both taken — it is now local to `usable_beam_mask` and the threshold is `np.finfo(np.float32).tiny`. 49f0db9.

You are right that the hand-picked value was worse: I had it at 1e-6 arcsec on the reasoning that `radio_beam`'s `Beam.__eq__` treats anything within 1e-10 deg of zero as a zero beam, so the sentinel's own unit would not matter. That is defensible but it hard-codes another library's tolerance to guard a case that does not arise — `fitscube` writes BMAJ/BMIN sentinels in arcsec, which is the unit compared in. Naming the actual constant is clearer.

On `> 0`: enough for the crash, not enough for the rest. The two placeholders in play are wsclean's exact zero and `fitscube`'s tiny, and `> 0` only catches the first. Since it is `> 0` plus the finite checks, it exactly matches `Beams.isfinite`, which is the mask `radio_beam` applies before its `argmax` — so it does guarantee the solver never sees an empty set. But it would let the tiny sentinel through, and then `cubes_share_common_beam` reads a blanked channel as a beam mismatch and reconvolves a cube that is already at one resolution. Which is most likely how this ran on real racs-all output in the first place.

---

PR #415, opened 2026-09-02
Measure the polarisation noise with BANE, and feed it to rm-synthesis

AlecThomson:
Two commits: the BANE port, then the wiring. Worth reviewing separately.

## Why

The racs-all smoke test produced all-zero moment maps and all-NaN peak maps, and the linmos weight cubes are the prime suspect for the noise estimate behind them. This replaces that inferred noise with one measured off the co-added planes themselves.

## 1. `flint/bane.py`, the port

Ported from `AegeanTools.BANE_fft` (AlecThomson/Aegean, `dask` branch). Only the 2D single-plane routines: the cube handling, chunking, `estimate_rms` (which needs `AegeanTools.numba_polyfit`) and the CLI all drop out.

`numba` and `rocket-fft` are new dependencies. The latter is not optional: numba's nopython mode has no `np.fft`, so without it every `njit` here fails to compile, eagerly at import for the three carrying explicit signatures. I verified that directly before adding it.

### Upstream bugs fixed, each measured on a synthetic sky

| | before | after |
|---|---|---|
| DC background of 0.01 on 1e-3 noise | 100% of pixels flagged as sources, background destroyed | recovers 9.99e-3 of 1e-2 |
| background at a 150 px blanked border | 3.59e-4, 18% of truth | 1.91e-3, 96% of truth |
| rms at that border | 9.39e-4, 94% of truth | 9.65e-4, 96% of truth |

1. **`get_kernel`'s beam guard never fired.** It caught `ValueError`; `radio_beam` raises `NoBeamException`, which inherits straight from `Exception`. A header with no beam escaped the friendly error entirely.
2. **A kernel wider than the image read out of bounds.** `pad_reflect` pads by the kernel shape and is `njit`-ed without bounds checking, so it returned quietly wrong maps rather than raising. A 256 px image with a 10 px beam downsamples to 6x6 against a 9x9 kernel and gave an RMS of exactly zero. Now refused with a message naming the fix.
3. **Sources were clipped on `|image| / rms`, not `|image - background| / rms`.** Any DC offset made every pixel a source.
4. **The clip threshold was one `mad_std` for the whole plane** while the noise it clips against is not uniform. Now two passes, the second clipping against the first pass's maps, so the cut follows the noise. This is the case the polarisation mosaics are bought for.
5. **Blanked pixels entered the convolution as measured zeros** while the normalisation still divided by the whole kernel, collapsing the background toward zero within a kernel width of any blank. Now normalised by the smoothed validity mask.

### Not fixed

The offset upstream records as a TODO. I tried sampling half a cell in, measured it, and it made the background *worse* (mean abs error 3.06e-4 against 2.68e-4), so the grid is left as upstream has it. My current read of the cause is in a comment on `_downsample_slices`: the sampled region spans `step` to `length - step` while the zoom back up stretches it across the full image, so it is misregistered by construction. Not attempted here.

## 2. The wiring

- `linmos_channel_groups_to_cubes` optionally runs BANE on each co-added channel as it lands, in parallel with the other channels, and stacks the background and RMS planes into cubes the same way image and weight planes are stacked.
- It now returns a `LinmosCubes` NamedTuple instead of a list. `create_convolve_linmos_cubes` calls `.futures` on it, so the continuum flows are untouched.
- The racs-all flow hands the RMS cubes to rm-synthesis in place of the weight cubes, and spices the background and RMS cubes alongside the rest.
- A weight cube is `1/sigma**2`, an RMS cube is `sigma`. rm-lite takes either through the same argument and is told which by `noise_files_are_weight`, so `run_rmsynth_3d` gains explicit noise-cube arguments, refuses both sets at once, and sets the flag accordingly.
- BANE is opt-in via `PolFieldOptions.bane_noise`; the flow falls back to the linmos weights when it was not run.
- `FFTBANEOptions` lives in `flint/options.py`, not `flint/bane.py`, so strategy validation does not import numba. Same reason `RMSynthOptions` lives there rather than in `flint/rmsynth.py`.

## Known limitation, worth deciding before merge

BANE runs on the linmos planes **before** cube assembly, while rm-synthesis convolves the cubes to a common beam before building the FDF. The RMS cubes therefore describe the pre-convolution resolution, which is the same mismatch the weight cubes have. Swapping weights for RMS fixes the scale and the spatial variation of the noise, but not this. Options discussed: measure after the convolution inside the rm-synth stage, accept the mismatch, or convolve before co-adding.

There is also an open question about whether the weight-cube path should be dropped from the flow entirely rather than kept as a fallback.

## Tests

`tests/test_bane.py` is new: `pad_reflect` against `np.pad`, a flat image through `fft_average`, recovery of a varying background and RMS, blanked pixels staying blank, seed reproducibility, beam-derived kernel sizing, and one test per fixed bug. `tests/test_racs_all_pipeline.py` gains three covering RMS cubes preferred over weights, the weight fallback, and the BANE cubes reaching SPICE.

Full suite was still running when this was opened; `test_bane.py`, `test_rmsynth.py` and `test_racs_all_pipeline.py` all pass, and ruff is clean. I will report the full result on the PR.

https://claude.ai/code/session_01MKV4YVPViaNXK1H6euKLhA

---
_Generated by [Claude Code](https://claude.ai/code/session_01MKV4YVPViaNXK1H6euKLhA)_

### DIFF
--- a/flint/options.py
+++ b/flint/options.py
@@ -231,7 +231,7 @@ class PolFieldOptions(BaseOptions):
     pb_cutoff: float = 0.1
     """Primary beam attenuation cutoff to use during linmos"""
     bane_noise: FFTBANEOptions | None = None
-    """Opt in to measuring a background and RMS cube off the co-added planes with BANE (see ``flint.bane``). These describe the natural-resolution cubes this stage archives. None, the default, skips it. Independent of ``RMSynthFieldOptions.bane_noise``, which is the one the FDF noise comes from"""
+    """Measure a background and RMS cube off the co-added planes with BANE. Off by default, and independent of ``RMSynthFieldOptions.bane_noise``"""
     imaging_strategy: Path | None = None
     """Path to a FLINT imaging yaml file that contains settings to use throughout imaging"""
     pol_cube_channel_width: float | None = None
@@ -241,13 +241,8 @@ class PolFieldOptions(BaseOptions):
 
 
 class FFTBANEOptions(BaseOptions):
-    """Options for ``flint.bane.robust_bane``. Named apart from
-    ``flint.source_finding.aegean.BANEOptions``, which drives the containerised
-    aegean BANE instead.
-
-    Defined here rather than in ``flint.bane`` so that strategy validation
-    (``flint.configuration``) does not pull in numba, as ``RMSynthOptions`` does
-    for rm_lite."""
+    """Options for ``flint.bane.robust_bane``. Lives here, not in ``flint.bane``,
+    so strategy validation does not pull in numba"""
 
     step_size: int | None = None
     """Downsampling factor in pixels. None uses 3 beams; a negative value sets the beams per step"""
@@ -256,31 +251,22 @@ class FFTBANEOptions(BaseOptions):
     clip_sigma: float = 5.0
     """Pixels above this SNR are replaced by noise before the background is fitted"""
     seed: int = 1234
-    """Seed for the noise the clipped pixels are filled with, so a rerun reproduces the maps"""
+    """Seeds the noise clipped pixels are filled with, so a rerun reproduces the maps"""
 
 
 class _StokesTrio(BaseOptions):
-    """One path per Stokes, shared by the containers below.
-
-    Never annotate with this directly: the whole point of the three concrete
-    types is that a cube cannot be mistaken for a noise or a weight, and a base
-    they all satisfy would let exactly that through.
-    """
+    """One path per Stokes. Never annotate with this: the types below exist to be told apart"""
 
     q: Path
     """Stokes Q"""
     u: Path
     """Stokes U"""
     i: Path | None = None
-    """Stokes I. Optional throughout: rm-synthesis runs without the fractional-polarisation correction"""
+    """Stokes I, optional: rm-synthesis runs without it"""
 
     @classmethod
     def from_mapping(cls, cubes: Mapping[str, Path]) -> Self:
-        """From a per-Stokes dict, as the polarisation stage keys them.
-
-        Anything but q/u/i is dropped, and a missing q or u is refused here
-        rather than surfacing as a KeyError from the call site.
-        """
+        """From a per-Stokes dict, dropping anything but q/u/i"""
         missing = {"q", "u"} - cubes.keys()
         if missing:
             msg = f"Need a cube for every one of q and u, missing {sorted(missing)}."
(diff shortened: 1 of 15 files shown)


flint/options.py
```diff
@@ -237,6 +240,87 @@ class PolFieldOptions(BaseOptions):
...
+
+
+StokesErrorCubes: TypeAlias = Annotated[
+    StokesWeightCubes | StokesNoiseCubes, Field(discriminator="kind")
+]
+"""The error cubes rm-synthesis weights by: an inverse variance or a noise, never
+both and never something ambiguous. rm-lite takes either through one argument and
+is told which by a flag, so mixing them inverts the noise by 1/sigma**2. The
+discriminator is what keeps that straight across a serialisation round trip, which
+prefect does to every task input."""
```
AlecThomson: @claude - this is too much text. Simplify

flint/options.py
```diff
@@ -349,18 +433,12 @@ class RMSynthFieldOptions(BaseOptions):
...
-    stokes_u_weight_cube: Path | None = None
-    """Path to Stokes U weights cube produced by LINMOS"""
-    stokes_i_weight_cube: Path | None = None
-    """Path to Stokes I weights cube produced by LINMOS"""
+    stokes_cubes: StokesCubes | None = None
+    """The Stokes Q/U (and optionally I) FITS cubes. Computed by the racs-all flow, so required only when running this pipeline standalone"""
+    error_cubes: StokesErrorCubes | None = None
+    """The linmos weight cubes or a set of noise cubes, never both: see ``StokesErrorCubes``. None leaves rm-lite to estimate a per-channel noise from Q/U itself. Superseded by ``bane_noise``, which measures the noise at the resolution the FDF is built at"""
+    bane_noise: FFTBANEOptions | None = None
+    """Opt in to measuring BANE background and RMS cubes off the common-resolution cubes this stage writes, and using the RMS for the FDF noise (see ``flint.bane``). Measured after the convolution, so they describe the resolution the FDF is built at and supersede ``error_cubes``, which describe the unconvolved inputs. None, the default, falls back to whatever cubes the caller passed"""
```
AlecThomson: @claude - docstrings must be terse one-liners. These are novels

AlecThomson (commented):
@claude - this is looking good. Once again you write too much text.

Question, though - we have mode `_StokesTrio` but will we ever have need for Stokes V? It 'feels' missing, but not sure if it'll be needed. Lets check for this.

flint/options.py
AlecThomson: Cut to two lines in `3511d80`:

```python
"""A weight or a noise, never both. Confusing them inverts the noise by 1/sigma**2,
so the tag keeps them apart across the serialisation prefect does to task inputs."""
```

Took this as applying to the whole branch rather than just here, so `bane.py`, `rmsynth.py` and the flow wiring got the same pass. 124 lines of prose out, 69 in.

---

flint/options.py
AlecThomson: All one-liners now, in `3511d80`:

```python
stokes_cubes: StokesCubes | None = None
"""The Stokes cubes. Set by the racs-all flow"""
error_cubes: StokesErrorCubes | None = None
"""Weight or noise cubes, never both. None lets rm-lite estimate from Q/U itself"""
bane_noise: FFTBANEOptions | None = None
"""Measure BANE cubes off the common-resolution cubes and take the FDF noise from the RMS. Supersedes ``error_cubes``, which describe the unconvolved inputs"""
```

`_StokesTrio` went from 6 lines to 1 and `from_mapping` from 5 to 1 in the same pass.

One judgement call worth flagging: `bane_noise` keeps its second clause. That it supersedes `error_cubes` is the non-obvious precedence rule, and the two are easy to confuse. Happy to cut it if you'd rather it were strictly one line.

---

PR #418, opened 2026-09-02
Take the rm-synth cubes as a list, so the standalone pipeline can run

AlecThomson:
## The problem

`flint_flow_rmsynth_pipeline` could not be run on its own at all. Passing the Stokes cubes in a `--cli-config` file failed with:

```
flint_flow_rmsynth_pipeline: error: --stokes-cubes can't be set to a list '[...q..., ...u..., ...i...]'
unless its action type is changed to 'append' or nargs is set to '*', '+', or > 1
```

`RMSynthFieldOptions.stokes_cubes` is a nested Options class (`CubesForRMSynth`), and `add_options_to_parser` renders anything non-iterable as a single-value `store` argument. So:

- configargparse rejects a list outright, which is the error above.
- Even a single string would fail, as pydantic will not coerce `str` into `CubesForRMSynth`.
- `error_cubes` is worse: a discriminated `WeightCubesForRMSynth | NoiseCubesForRMSynth` union that no one CLI string can express.

The `racs-all` flow never hit this. It already excludes both fields from its combined CLI (`COMPUTED_FIELDS`) and sets them in memory from the polarisation stage.

## The fix

Keep the model-typed fields as the programmatic API, so the `racs-all` handoff keeps its exact per-Stokes mapping, and give the standalone CLI a flat surface:

- Exclude `stokes_cubes`/`error_cubes` from the generated parser and add them by hand as `nargs="+"` lists of paths.
- Which Stokes each cube is comes from its filename, via the existing `split_images(by="pol")`, in a new `_CubesForRMSynth.from_paths` classmethod beside `from_mapping`.
- `--error-cubes` requires a new `--error-cube-kind {weight,noise}`. It is required rather than guessed because reading a weight as a noise inverts the error by 1/sigma**2.
- `_exclude_fields_cli_class` moves from `racs_all_pipeline` to `flint.options` as `exclude_fields_cli_class` so both flows can use it; importing it the other way round would be circular.

The originally failing config now works unchanged:

```
stokes-cubes = [SB56289.RACS_1041+18.round1.q.pol.image.cube.fits, SB56289.RACS_1041+18.round1.u.pol.image.cube.fits]
```

## Notes

Recovering the Stokes from the filename means cube names have to follow the `flint` or CASDA scheme, which the polarisation pipeline already requires of its inputs. Both a name that does not parse and one carrying no polarisation raise a message saying so, rather than failing later inside rm-synthesis.

## Tests

Nine tests in `tests/test_prefect_rmsynth_flow.py`: the failing `--cli-config` case, the equivalent command line, Stokes I being optional, `--error-cube-kind` picking the container, `cli()` assembling the nested options, `--error-cubes` without a kind, and the missing-Stokes / duplicate-Stokes / unparseable-name errors.

`pre-commit run --hook-stage manual --all-files` is clean and the suite is 544 passed. Three failures are environmental in this sandbox and unrelated to the diff: two `test_catalogue` tests cannot reach `vizier.cds.unistra.fr` through the proxy, and `test_copy_directory` needs `rsync`, which is not installed here.

Also adds a short `Running RM-synthesis on its own` section to the polarisation quickstart, which documented no rm-synth CLI before.

https://claude.ai/code/session_01LWfbjm7f9FrUXuEiMmZQab

---
_Generated by [Claude Code](https://claude.ai/code/session_01LWfbjm7f9FrUXuEiMmZQab)_

### DIFF
--- a/docs/quickstart/polarisation.md
+++ b/docs/quickstart/polarisation.md
@@ -61,44 +61,6 @@ cubes alone, so both are available: the natural cubes to archive, the total cube
 to synthesise from. See `RMSynthFieldOptions.beam_cutoff` to drop the coarsest
 channels from that solve rather than smoothing the whole band to reach them.
 
-## Running RM-synthesis on its own
-
-`flint_flow_rmsynth_pipeline` runs the rm-synth stage against cubes that have
-already been imaged, without the rest of the `racs-all` flow. The Stokes cubes
-are given as a flat list, and which Stokes each one is comes from its filename,
-so they have to follow the `flint` or CASDA naming scheme:
-
-```bash
-flint_flow_rmsynth_pipeline \
-  --stokes-cubes SB56289.RACS_1041+18.round1.q.pol.image.cube.fits \
-                 SB56289.RACS_1041+18.round1.u.pol.image.cube.fits \
-                 SB56289.RACS_1041+18.round1.i.pol.image.cube.fits \
-  --imaging-strategy strategy.yaml
-```
-
-Stokes Q and U are required; Stokes I is optional and is used to fit the
-fractional-polarisation correction. `--error-cubes` takes a matching list, and
-needs `--error-cube-kind` to say whether those are `linmos` weights (an inverse
-variance) or a noise (a sigma) -- reading one as the other inverts the error by
-1/sigma**2. Every option can equally be set in a `--cli-config` file:
-
-```
-stokes-cubes = [SB56289.RACS_1041+18.round1.q.pol.image.cube.fits, SB56289.RACS_1041+18.round1.u.pol.image.cube.fits]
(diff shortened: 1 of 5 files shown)


docs/quickstart/polarisation.md
```diff
@@ -61,6 +61,44 @@ cubes alone, so both are available: the natural cubes to archive, the total cube
...
+```{argparse}
+:ref: flint.prefect.flows.rmsynth_pipeline.get_parser
+:prog: flint_flow_rmsynth_pipeline
+```
+
+## The `RMSynthFieldOptions` class
+
+```{literalinclude}  ../../flint/options.py
+:pyobject: RMSynthFieldOptions
+```
```
AlecThomson: @claude - not needed for now

AlecThomson: Dropped in 06cf0f4. The whole section is gone (usage examples, the `argparse` directive and the `RMSynthFieldOptions` literalinclude), so this PR no longer touches `docs/` at all.

---

PR #420, opened 2026-09-06
Align the FFT BANE maps with the plane they measure, and halve their memory

AlecThomson:
Started as an investigation into `bane-fft` tasks appearing to hold memory in the Dask dashboard. That turned out to be a misreading, but looking closely at the routine turned up a real alignment bug, which is now the important half of this.

**The background and noise maps came out displaced up and to the right of the plane they describe.** Anything already made with the current code carries that shift, and anything derived from those maps — source-finding thresholds, the inverse-variance weights rm-synthesis takes from the RMS cube — inherits it. Those products want regenerating.

### 1. `efb9b1e` — working memory, 9.5x the plane to 4.3x

`bane_fits_image` runs as one task per channel, so what a worker has to fit is the peak `robust_bane` reaches, not the two paths it hands back. Six full-resolution arrays that did not need to exist:

- the seed background and noise were two whole arrays holding a single number each — now scalars
- the compacted copy for `median`/`mad_std` was held for the whole routine — now dropped before the rounds it seeds
- the clip mask built three full-size temporaries — now in place
- the validity mask was full resolution but only ever consumed downsampled — now built downsampled
- the zero-filled copy of the plane was held across both rounds — each round now builds its clipped plane from the image directly
- `nan_to_num` and the negative-noise clip copied rather than working in place, and the source count was summed three times

Measured by RSS: 588 MB to 287 MB at 4000x4000, 1287 MB to 588 MB at 6000x6000.

**Output is bit-identical** across fourteen cases — NaN and linmos-zero blanking, a wholly blank plane, a plane with no beam, DC offsets, gradients, fixed and beam-derived kernel sizes, no downsampling, and both clip and seed settings. 28/28 arrays, worst difference 0.

### 2. `939a1ad` — the alignment bug

Two causes, both displacing the maps the same way, so they added.

**The kernel was never centred.** `_ft_kernel` zero-pads out from index zero rather than centring on the origin, so every `fft_average` convolves and translates in one go, by half the kernel. Convolving a delta shows it exactly: a (13, 13) kernel moved it 6 pixels along each axis, a (19, 19) kernel 9. Because the kernel lives on the downsampled grid, half a kernel is `step * box//2` in plane pixels — tens of pixels at the default sizes. Taking the output window at that displacement costs nothing and now matches `ndimage.convolve` to 1e-6.

`test_fft_average_preserves_a_flat_image` could never have caught this: a flat image is invariant under translation.

**The step back up scaled the maps over the whole plane**, while the samples that made them start a step in and stop a step short. `ndimage.zoom` takes a scale and nothing else, so it could not know. `_to_full_resolution` uses an affine transform, which carries the offset too. Recovering a ramp through the round trip, mean error over the sampled region falls from 10.7-21.1 pixels to 0.01-0.14, and it holds when the step does not divide the plane — which resampling half a cell in, the other candidate fix, does not.

End to end on a 512x512 plane with a blob of louder noise, cross-correlated against a centred reference:

| | before | after |
|---|---|---|
| x | +85.0 px | -3.7 px |
| y | +93.5 px | +7.6 px |

The residual is the probe's own precision — the sign differs between axes.

### 3. `306a40c` — linear interpolation instead of a cubic spline

A blank leaves a step in the map, where `nan_to_num` has put zero against a measured value, and the cubic overshoots it. On a 2000x2000 plane blanked to an eighth, the background **that was written out** reached -9.67 against an image whose noise is 1e-3, with the worst pixel one row past the blank edge. The RMS undershot to a floor of exactly zero — the clip firing, not a measured value.

| | cubic | linear |
|---|---|---|
| delivered background range | [-9.67, +4.49] | [-1.1e-3, +8.9e-4] |
| RMS floor | 0.000 (clamped) | 6.2e-4 (real) |
| time, 2000x2000 | 1.86 s | 0.81 s |

Away from the blanks the two agree to well under a percent of the RMS, so what is given up is smoothness that was never resolved — the samples are a step apart. The clip on negative noise stays, now as a guard rather than a fix, so raising the order again cannot quietly deliver a negative error.

### Follow-ups on review

`4fa9943` cuts the commentary back and inlines `_seed_estimates` and `_where_masked` — ninety lines out, thirty in, maps bit-identical. `1794b59` fixes a `typos` hit CI caught (`codespell` passes the same string, which is how it got through) and a docstring the linear step had made stale. `fdf8368` covers the `step_size=0` branch that Codecov flagged.

## Tests

Five new ones, each of which fails before the change it covers, or exercises a path nothing else did:

- `test_fft_average_puts_the_smoothed_pixel_over_the_pixel_it_smooths` — a delta against a centred convolution
- `test_the_noise_map_lines_up_with_the_noise_it_measures` — a blob of louder noise found within ten pixels of where it was put, against sixty-odd before
- `test_the_working_memory_stays_a_small_multiple_of_the_plane` — peak under 7x the plane, against 9.5x before
- `test_a_wider_plane_is_measured_as_float32` — pins the dtype contract
- `test_a_plane_measured_without_downsampling` — the one path that never steps the maps back up, so the only end-to-end check of the kernel centring on its own

The seventeen existing tests pass unchanged. None of them covered where a map puts a feature, which is why this ran for as long as it did.

`flint/bane.py` reads as 86% covered because `coverage` cannot see inside a compiled `njit` function; every uncovered line is inside one.

## Worth a second opinion

- **The memory bound is a judgement call.** 7x against 4.3-5.4x measured locally; numpy and scipy allocate differently between versions and CI spans 3.11 to 3.13. It has since passed on 3.13, so that is one version of evidence rather than three. If it flakes, loosen it rather than drop it.
- **The kernel centring is probably upstream too.** This was ported from `AegeanTools.BANE_fft` and the uncentred `rfft2(kernel, s=shape)` looks like it came across with it. Someone who knows that codebase should check.
- `robust_bane` now casts the plane to float32 up front, which it had been getting incidentally from the zero-filled copy that has gone.

https://claude.ai/code/session_018Skuoh52Dn8WdnNmvijZvR

### DIFF
--- a/flint/bane.py
+++ b/flint/bane.py
@@ -91,13 +91,6 @@ def fft_average(
 
     Reflect-padded by the kernel size so the FFT's periodic wrap does not fold
     the far edge of the image back onto the near one.
-
-    ``_ft_kernel`` zero-pads the kernel out from index zero rather than centring
-    it on the origin, which convolves and displaces in one go: the result comes
-    back half a kernel further along each axis. The window is taken at that
-    displacement, which costs nothing and leaves a smoothed pixel sitting over
-    the pixel it smooths - a flat image hides this, but a noise map laid over
-    the image it describes does not.
     """
     pad_x, pad_y = kernel.shape
     image_padded = pad_reflect(array=image, pad_width=(pad_x, pad_y))
@@ -106,6 +99,8 @@ def fft_average(
     kernel_fft = _ft_kernel(kernel, shape=image_padded.shape)
     smooth = fft.irfft2(image_fft * kernel_fft, s=image_padded.shape) / kernel.sum()
 
+    # `_ft_kernel` pads from index zero, not about the origin, so the result
+    # comes back half a kernel along each axis
     nx, ny = image.shape
     start_x, start_y = pad_x + pad_x // 2, pad_y + pad_y // 2
     return smooth[start_x : start_x + nx, start_y : start_y + ny]
@@ -220,12 +215,8 @@ def _downsample_slices(
 ) -> tuple[slice, slice]:
     """Slices taking every `step_size_pix` pixel, trimmed to an even count.
 
-    The sampled region runs from `step_size_pix` to `length - step_size_pix`,
-    so it does not cover the plane. `_to_full_resolution` puts the maps back
-    where these slices took them from, which is what upstream's TODO about an
-    offset is: scaling them over the whole plane instead moves them bodily.
-    Sampling half a cell in so a plain scaling would be honest was tried and is
-    only exact when the step divides the plane, so this keeps upstream's grid.
+    These do not cover the plane, so `_to_full_resolution` reads them to put
+    the maps back where they were taken from.
     """
     slices = []
     for length in (shape[0], shape[1]):
@@ -243,24 +234,9 @@ def _to_full_resolution(
 ) -> NDArray[np.float32]:
     """Put a map measured on the downsampled grid back onto the plane's own grid.
 
-    ``ndimage.zoom`` takes a scale and nothing else, and the sampled grid is not
-    a pure scaling of the plane's: it starts a step in and stops a step short.
-    Stretching the maps over the whole plane instead of over the region actually
-    sampled moves every pixel of them - up and to the right, by tens of pixels
-    for a typical step, and growing across the field, since the error is a
-    stretch as well as a shift. The affine transform carries the offset too, so
-    a sample lands back on the pixel it was taken from.
-
-    Outside the sampled region the nearest sample is held rather than the grid
-    reflected, so the extrapolated border does not ring.
-
-    Linear rather than a cubic spline. A blank leaves a step in the map, where
-    ``nan_to_num`` has put zero against a measured value, and a cubic overshoots
-    it: a plane blanked to an eighth came back with a background reaching nine
-    thousand times its own noise a single row past the blank edge, in the map
-    that was written out. Linear cannot overshoot, being a weighted mean of the
-    two samples either side, and away from the blanks it differs from the cubic
-    by well under a percent of the noise - for half the time.
+    An affine transform rather than `ndimage.zoom`, which takes a scale alone
+    and so cannot know the sampled grid starts a step in. Linear rather than a
+    spline, which overshoots the step a blank leaves in the map.
 
     Args:
         smoothed (NDArray[np.float32]): A map on the downsampled grid
@@ -282,17 +258,6 @@ def _to_full_resolution(
         order=1,
         mode="nearest",
     )
-
-
-def _where_masked(
-    value: NDArray[np.float32] | np.float32, mask: NDArray[np.bool_]
-) -> NDArray[np.float32] | np.float32:
-    """``value`` at the set pixels of `mask`, taking a scalar to mean everywhere.
-
-    Round one clips against one number for the whole plane, round two against
-    the maps round one made, and the refill below reads both the same way.
-    """
-    return value[mask] if isinstance(value, np.ndarray) else value
 
 
 def _bane_round(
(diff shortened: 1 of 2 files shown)


flint/bane.py
```diff
@@ -91,6 +91,13 @@ def fft_average(
 
     Reflect-padded by the kernel size so the FFT's periodic wrap does not fold
     the far edge of the image back onto the near one.
+
+    ``_ft_kernel`` zero-pads the kernel out from index zero rather than centring
+    it on the origin, which convolves and displaces in one go: the result comes
+    back half a kernel further along each axis. The window is taken at that
+    displacement, which costs nothing and leaves a smoothed pixel sitting over
+    the pixel it smooths - a flat image hides this, but a noise map laid over
+    the image it describes does not.
```
AlecThomson: @claude - commentary not needed here

flint/bane.py
```diff
@@ -211,10 +220,12 @@ def _downsample_slices(
...
-    The sampled region runs from `step_size_pix` to `length - step_size_pix`
-    while the zoom back up stretches it over the whole image. That is the likely
-    cause of the offset upstream records as a TODO; sampling half a cell in
-    instead measured worse, so this keeps upstream's grid.
+    The sampled region runs from `step_size_pix` to `length - step_size_pix`,
+    so it does not cover the plane. `_to_full_resolution` puts the maps back
+    where these slices took them from, which is what upstream's TODO about an
+    offset is: scaling them over the whole plane instead moves them bodily.
+    Sampling half a cell in so a plain scaling would be honest was tried and is
+    only exact when the step divides the plane, so this keeps upstream's grid.
```
AlecThomson: @claude commentary not needed

flint/bane.py
```diff
@@ -225,12 +236,70 @@ def _downsample_slices(
...
+    Outside the sampled region the nearest sample is held rather than the grid
+    reflected, so the extrapolated border does not ring.
+
+    Linear rather than a cubic spline. A blank leaves a step in the map, where
+    ``nan_to_num`` has put zero against a measured value, and a cubic overshoots
+    it: a plane blanked to an eighth came back with a background reaching nine
+    thousand times its own noise a single row past the blank edge, in the map
+    that was written out. Linear cannot overshoot, being a weighted mean of the
+    two samples either side, and away from the blanks it differs from the cubic
+    by well under a percent of the noise - for half the time.
```
AlecThomson: @claude 
```suggestion
```

flint/bane.py
```diff
@@ -241,62 +310,119 @@ def _bane_round(
...
-        # obviously wrong
-        rms = np.clip(rms, 0.0, None)
+        smooth_rms = _to_full_resolution(smooth_rms, sampled_at, full_shape)
+        # A guard rather than a fix: the linear interpolation above is a
+        # weighted mean of two non-negative samples and so cannot undershoot.
+        # It is what makes the interpolation order safe to raise again, since a
+        # spline does undershoot across the step `nan_to_num` leaves at the
+        # footprint edge - and a negative error squares to a small positive
+        # variance, so an inverse-variance weight downstream would come out
+        # orders of magnitude too large rather than obviously wrong
```
AlecThomson: @claude 
```suggestion
```

flint/bane.py
```diff
@@ -241,62 +310,119 @@ def _bane_round(
...
+        rms_estimator (Callable): Estimator of the plane's noise, e.g. ``mad_std``
+
+    Returns:
+        tuple[np.float32, np.float32] | None: Seed background and RMS, or None if the plane is wholly blank
+    """
+    finite = image[~nan_mask].ravel()
+    if finite.size == 0:
+        return None
+    # float32, matching the maps they seed and the plane they came from
+    return np.float32(np.median(finite)), np.float32(rms_estimator(finite))
```
AlecThomson: @claude - seems like these could be in-placed, mostly? Otherwise, we have a novel of docs for 4 lines of actual funciton

flint/bane.py
AlecThomson: Cut in 4fa9943. The paragraph is gone; two lines sit at the slice itself, since the offset looks like an off-by-one and would otherwise invite a revert.

---

flint/bane.py
AlecThomson: Cut in 4fa9943, down to one sentence pointing at `_to_full_resolution`.

---

flint/bane.py
AlecThomson: Deleted in 4fa9943 — three paragraphs down to three lines saying what it does differently from `ndimage.zoom` and why linear.

---

flint/bane.py
AlecThomson: Deleted in 4fa9943. Kept one sentence rather than nothing, since the `np.clip` is a no-op as the code stands and reads like dead code without it.

---

flint/bane.py
AlecThomson: Agreed — inlined in 4fa9943. `_seed_estimates` is gone; the median and mad_std now sit in `robust_bane` with a `del finite` where the helper's return had been dropping the compacted copy. Same memory behaviour, explicit rather than a side effect of scope.

I took `_where_masked` the same way, since it was one line of logic behind a docstring — it's now a branch on the two values together at the one place that used it. Say if you'd rather have that one back.

Ninety lines out, thirty in, and the maps are bit-identical across the fourteen cases I've been checking against.

---

PR #421, opened 2026-09-06
Update rm-lite to v2026.9.1 and write its new Faraday moment products

AlecThomson:
Bumps `rm-lite` to [v2026.9.1](https://github.com/AlecThomson/rm-lite/releases/tag/v2026.9.1) and wires in what it now returns.

## The new moment products

`FaradayMoments` gained an error for each moment, a debiased `mom0`, and the polarisation at the reference lambda^2. Those are the maps the moment set was missing: `mom0` alone says how much polarised flux a pixel integrates but nothing about whether it is a detection, and the peak maps have carried their errors for a while.

`_MOMENT_MAPS` replaces `_MOMENT_NAMES`, taking the same `{field: (suffix, BUNIT, comment)}` shape as `_PEAK_MAPS`. Twelve maps per requested FDF now:

| suffix | BUNIT | |
| --- | --- | --- |
| `mom0`, `mom0_debias`, `mom0_error` | Jy/beam | zeroth moment, less the noise's own sum, and its 1-sigma |
| `mom1`, `mom1_error` | rad/m2 | mean Faraday depth |
| `mom2`, `mom2_error` | rad/m2 | Faraday depth dispersion |
| `pi_lam_sq_0`, `pi_lam_sq_0_debias`, `pi_lam_sq_0_error` | Jy/beam | polarised intensity at lambda^2_0 |
| `pa_lam_sq_0`, `pa_lam_sq_0_error` | deg | polarisation angle there |

Each map is stamped with its unit. The moments used to be three maps that all read as fluxes; Jy/beam, rad/m2 and degrees now share the one set, so an unstamped map is easy to misread.

Two details worth flagging:

- The theoretical FDF noise is passed to `calc_faraday_moments`. rm-lite defaults `fdf_error` to `None`, and without it *every* error and debiased map comes back NaN rather than failing — so a test asserts they are actually measured.
- The `.debiased` set (`RMSynthOptions.debias_moments`) skips `mom0_debias`. `debias_fdf` already takes the noise off each amplitude, which is exactly what `mom0_debias` subtracts, so rm-lite returns `mom0` unchanged there and the map would be a byte-identical copy of the already-written one under a second name.

`pi_lam_sq_0` is only meaningful against a single reference wavelength, so `RMSynthOptions.lam_sq_0_m2` now notes that `per_pixel` leaves those two maps measuring every pixel at a different lambda^2.

## Dropping the peak SNR cut

rm-lite removed `threshold` from `calc_faraday_peaks`: the peak and its error are both reported now, so `peak_pi / peak_pi_error` is the SNR to select on afterwards. `peak_threshold_snr` goes with it, from `RMSynthFieldOptions` through `write_rm_products` and the prefect task.

It defaulted to `0.0` (no cut), so this changes nothing for a default run. **`BaseOptions` sets `extra: "forbid"`, so a config or strategy still setting `peak_threshold_snr` will now fail validation** rather than silently applying no cut. Nothing in this repo sets it. The moment cut (`moment_threshold_snr`) is untouched and still needed — `mom0` integrates the whole Faraday depth axis, so an uncut off-source pixel piles hundreds of noise samples into a positive floor.

## A latent bug the new maps exposed

`_compute_rm_products` keyed its results by `Future`. Products built from the same expression share a dask key, so `client.compute` hands back **one** future for all of them — and rm-lite derives `pi_lam_sq_0_error` exactly as it derives `mom0_error`. `dict(zip(futures, compute_keys))` then dropped all but the last key of each such group, surfacing as a `KeyError` on a map that had in fact been computed. The mapping is one-to-many now, and `test_products_sharing_a_dask_key_are_all_returned` pins it against a distributed client without depending on rm-lite continuing to build those two the same way.

## Testing

`tests/test_rmsynth.py`, `tests/test_prefect_rmsynth_flow.py`, `tests/test_racs_all_pipeline.py` and `tests/test_options.py` all pass (94 passed). Tests that counted moment maps by hand now count off `_MOMENT_MAPS`, so the next rm-lite moment does not need them edited again. Ruff clean; `mypy flint/rmsynth.py` reports only the pre-existing untyped-import and zarr `store` complaints.

https://claude.ai/code/session_018Skuoh52Dn8WdnNmvijZvR

---
_Generated by [Claude Code](https://claude.ai/code/session_018Skuoh52Dn8WdnNmvijZvR)_

### DIFF
--- a/flint/options.py
+++ b/flint/options.py
@@ -374,13 +374,13 @@ class RMSynthOptions(BaseOptions):
     n_error_samples: int = 1000
     """Monte-Carlo resamples used by compute_model_error"""
     debias_moments: bool = False
-    """Also compute a debiased (via rm_lite's debias_fdf) moment set per requested FDF, written with a '.debiased' suffix. Distinct from the mom0_debias map, which takes the noise's own sum off mom0 rather than debiasing the FDF first"""
+    """Also compute a debiased (via rm_lite's debias_fdf) moment set per requested FDF, written with a '.debiased' suffix."""
     debias_filter_size: int = 5
     """Median filter size (pixels) used by mom0 debiasing"""
     lam_sq_0_m2: float | Literal["auto", "per_pixel"] = "auto"
-    """Reference lambda^2 the FDF is derotated to. 'auto' picks one value for the whole cube; 'per_pixel' gives each pixel its own, which also forces per_pixel_rmsf since the RMSF then differs pixel to pixel, and leaves the pi/pa_lam_sq_0 moment maps measuring every pixel at a different wavelength. A float pins it explicitly"""
+    """Reference lambda^2 the FDF is derotated to. 'auto' picks one value for the whole cube; 'per_pixel' gives each pixel its own, which also forces per_pixel_rmsf"""
     per_pixel_rmsf: bool = False
-    """Compute the RMSF for each pixel rather than one shared by the cube. Roughly doubles the FDF's size. rm-lite turns this on itself when the weights make pixels disagree, as the linmos weight cubes do"""
+    """Compute the RMSF for each pixel"""
     estimate_stokes_i_noise: bool = True
     """Derive the per-channel Stokes I error from the Stokes I cube when no Stokes I weight cube is given. A weight cube takes precedence"""
 
@@ -460,11 +460,11 @@ class RMSynthFieldOptions(BaseOptions):
     cube_products: list[Literal["dirty", "clean", "model"]] = []
     """Which Faraday dispersion function (FDF) cubes to write as FITS. Nothing by default, as these cubes can be large."""
     moment_products: list[Literal["dirty", "clean", "model"]] = ["clean"]
-    """Which FDF(s) to compute Faraday moment maps from: mom0 (raw, debiased and error), mom1 and mom2 with their errors, and the polarised intensity (raw, debiased and error) and angle (and error) at the reference lambda^2. Twelve (ny, nx) maps per FDF"""
+    """Which FDF(s) to compute Faraday moment maps from: 'dirty', 'clean', 'model'"""
     moment_threshold_snr: float = 5.0
     """SNR cut (times the theoretical FDF noise) applied before the moment maps are computed, the dirty ones included"""
     peak_products: list[Literal["dirty", "clean", "model"]] = []
-    """Which FDF(s) to measure peak statistics from: peak polarised intensity (raw and debiased), Faraday depth, polarisation angle and intrinsic angle, each with its error. Nine (ny, nx) maps per FDF, so empty by default -- at 16032^2 that is ~9 GB per entry. No SNR cut is applied: a peak is a single sample with no noise floor to integrate, so select on peak_pi / peak_pi_error afterwards"""
+    """Which FDF(s) to measure peak statistics from: 'dirty', 'clean', 'model'"""
     output_path: Path | None = None
     """Directory the FDF cube and moment products are written into. Defaults to alongside the input Stokes cubes"""
     sbid_copy_path: Path | None = None
--- a/flint/options.py
+++ b/flint/options.py
@@ -374,11 +374,11 @@ class RMSynthOptions(BaseOptions):
     n_error_samples: int = 1000
     """Monte-Carlo resamples used by compute_model_error"""
     debias_moments: bool = False
-    """Also compute a debiased (via rm_lite's debias_fdf) mom0/mom1/mom2 set per requested FDF"""
+    """Also compute a debiased (via rm_lite's debias_fdf) moment set per requested FDF, written with a '.debiased' suffix. Distinct from the mom0_debias map, which takes the noise's own sum off mom0 rather than debiasing the FDF first"""
     debias_filter_size: int = 5
     """Median filter size (pixels) used by mom0 debiasing"""
     lam_sq_0_m2: float | Literal["auto", "per_pixel"] = "auto"
-    """Reference lambda^2 the FDF is derotated to. 'auto' picks one value for the whole cube; 'per_pixel' gives each pixel its own, which also forces per_pixel_rmsf since the RMSF then differs pixel to pixel. A float pins it explicitly"""
+    """Reference lambda^2 the FDF is derotated to. 'auto' picks one value for the whole cube; 'per_pixel' gives each pixel its own, which also forces per_pixel_rmsf since the RMSF then differs pixel to pixel, and leaves the pi/pa_lam_sq_0 moment maps measuring every pixel at a different wavelength. A float pins it explicitly"""
     per_pixel_rmsf: bool = False
     """Compute the RMSF for each pixel rather than one shared by the cube. Roughly doubles the FDF's size. rm-lite turns this on itself when the weights make pixels disagree, as the linmos weight cubes do"""
     estimate_stokes_i_noise: bool = True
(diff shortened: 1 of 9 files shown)


flint/options.py
```diff
@@ -374,11 +374,11 @@ class RMSynthOptions(BaseOptions):
     n_error_samples: int = 1000
     """Monte-Carlo resamples used by compute_model_error"""
     debias_moments: bool = False
-    """Also compute a debiased (via rm_lite's debias_fdf) mom0/mom1/mom2 set per requested FDF"""
+    """Also compute a debiased (via rm_lite's debias_fdf) moment set per requested FDF, written with a '.debiased' suffix. Distinct from the mom0_debias map, which takes the noise's own sum off mom0 rather than debiasing the FDF first"""
```
```suggestion
    """Also compute a debiased (via rm_lite's debias_fdf) moment set per requested FDF, written with a '.debiased' suffix."""
```

flint/options.py
```diff
@@ -374,11 +374,11 @@ class RMSynthOptions(BaseOptions):
...
     debias_filter_size: int = 5
     """Median filter size (pixels) used by mom0 debiasing"""
     lam_sq_0_m2: float | Literal["auto", "per_pixel"] = "auto"
-    """Reference lambda^2 the FDF is derotated to. 'auto' picks one value for the whole cube; 'per_pixel' gives each pixel its own, which also forces per_pixel_rmsf since the RMSF then differs pixel to pixel. A float pins it explicitly"""
+    """Reference lambda^2 the FDF is derotated to. 'auto' picks one value for the whole cube; 'per_pixel' gives each pixel its own, which also forces per_pixel_rmsf since the RMSF then differs pixel to pixel, and leaves the pi/pa_lam_sq_0 moment maps measuring every pixel at a different wavelength. A float pins it explicitly"""
```
```suggestion
    """Reference lambda^2 the FDF is derotated to. 'auto' picks one value for the whole cube; 'per_pixel' gives each pixel its own, which also forces per_pixel_rmsf"""
```

flint/options.py
```diff
@@ -374,11 +374,11 @@ class RMSynthOptions(BaseOptions):
...
     per_pixel_rmsf: bool = False
     """Compute the RMSF for each pixel rather than one shared by the cube. Roughly doubles the FDF's size. rm-lite turns this on itself when the weights make pixels disagree, as the linmos weight cubes do"""
```
```suggestion
    """Compute the RMSF for each pixel"""
```

flint/options.py
```diff
@@ -460,13 +460,11 @@ class RMSynthFieldOptions(BaseOptions):
     cube_products: list[Literal["dirty", "clean", "model"]] = []
     """Which Faraday dispersion function (FDF) cubes to write as FITS. Nothing by default, as these cubes can be large."""
     moment_products: list[Literal["dirty", "clean", "model"]] = ["clean"]
-    """Which FDF(s) to compute Faraday moment maps from."""
+    """Which FDF(s) to compute Faraday moment maps from: mom0 (raw, debiased and error), mom1 and mom2 with their errors, and the polarised intensity (raw, debiased and error) and angle (and error) at the reference lambda^2. Twelve (ny, nx) maps per FDF"""
```
```suggestion
    """Which FDF(s) to compute Faraday moment maps from: ..."""
```
@claude then in the `...` just list the options

flint/options.py
```diff
@@ -460,13 +460,11 @@ class RMSynthFieldOptions(BaseOptions):
...
     moment_threshold_snr: float = 5.0
     """SNR cut (times the theoretical FDF noise) applied before the moment maps are computed, the dirty ones included"""
-    peak_threshold_snr: float = 0.0
-    """SNR cut (times the theoretical FDF noise) below which peak statistics are blanked. Zero applies no cut: a peak is a single sample with no noise floor to integrate, and peak_pi_error is written beside it"""
     peak_products: list[Literal["dirty", "clean", "model"]] = []
-    """Which FDF(s) to measure peak statistics from: peak polarised intensity (raw and debiased), Faraday depth, polarisation angle and intrinsic angle, each with its error. Nine (ny, nx) maps per FDF, so empty by default -- at 16032^2 that is ~9 GB per entry"""
+    """Which FDF(s) to measure peak statistics from: peak polarised intensity (raw and debiased), Faraday depth, polarisation angle and intrinsic angle, each with its error. Nine (ny, nx) maps per FDF, so empty by default -- at 16032^2 that is ~9 GB per entry. No SNR cut is applied: a peak is a single sample with no noise floor to integrate, so select on peak_pi / peak_pi_error afterwards"""
```
```suggestion
    """Which FDF(s) to measure peak statistics from: ..."""
```
@claude then replace `...` with just the possible options

flint/rmsynth.py
```diff
@@ -55,11 +55,36 @@
 
 FDFLabel: TypeAlias = Literal["dirty", "clean", "model"]
 FDFThreshold: TypeAlias = float | np.ndarray | da.Array | None
-_MOMENT_NAMES = ("mom0", "mom1", "mom2")
+
+# The Faraday moments, as {FaradayMoments field: (file suffix, BUNIT, comment)}.
+# The moments used to be three unitless-looking maps; they now carry an error
+# each, and the polarisation at the reference lambda^2, which mixes Jy/beam,
+# rad/m2 and degrees in the one set -- so every map is stamped with its BUNIT.
+_MOMENT_MAPS = {
```
AlecThomson:
```suggestion
MOMENT_MAPS = {
```

flint/rmsynth.py
```diff
@@ -55,11 +55,36 @@
 
 FDFLabel: TypeAlias = Literal["dirty", "clean", "model"]
 FDFThreshold: TypeAlias = float | np.ndarray | da.Array | None
-_MOMENT_NAMES = ("mom0", "mom1", "mom2")
+
+# The Faraday moments, as {FaradayMoments field: (file suffix, BUNIT, comment)}.
+# The moments used to be three unitless-looking maps; they now carry an error
+# each, and the polarisation at the reference lambda^2, which mixes Jy/beam,
+# rad/m2 and degrees in the one set -- so every map is stamped with its BUNIT.
```
AlecThomson:
```suggestion
```

flint/rmsynth.py
```diff
@@ -55,11 +55,36 @@
...
+}
+
+# ``debias=True`` rebuilds the FDF with the noise taken off each amplitude
+# first, which is exactly what rm-lite's own ``mom0_debias`` subtracts -- so it
+# returns ``mom0`` unchanged there, and writing a byte-identical copy of the
+# already-written ``mom0`` map under a second name would only invite reading
+# them as two measurements.
+_DEBIASED_MOMENT_MAPS = {
+    field: entry for field, entry in _MOMENT_MAPS.items() if field != "mom0_debias"
+}
```
AlecThomson: @claude - is this even needed?

flint/rmsynth.py
```diff
@@ -55,11 +55,36 @@
...
 
 # The FDF peak statistics, as {FaradayPeaks field: (file suffix, BUNIT, comment)}.
-# Units matter more here than for the moments: three of these are angles in
-# degrees and two are Faraday depths, so a map with no BUNIT is easy to misread.
 _PEAK_MAPS = {
```
AlecThomson:
```suggestion
PEAK_MAPS = {
```

flint/rmsynth.py
```diff
@@ -194,40 +219,38 @@ def run_rmclean_3d(
 
 
 def write_moment_maps_to_fits(
-    moments: FaradayMoments,
+    moments: dict[str, np.ndarray],
```
AlecThomson: @claude - you have replaced a nice Type with a dict

flint/rmsynth.py
```diff
@@ -558,20 +586,30 @@ def _compute_rm_products(
         futures = scheduler.compute(
             [compute_targets[key] for key in compute_keys], sync=False
         )
-    future_to_key = dict(zip(futures, compute_keys))
+    # Products that are the same expression share a dask key, so `compute`
+    # hands back one future for all of them -- rm-lite builds `mom0_error` and
+    # `pi_lam_sq_0_error` identically, for instance. One key per future would
+    # then quietly drop every product but the last of each such group, so the
+    # mapping is one-to-many and every key it carries is filled in.
```
AlecThomson: @claude - this is way too verbose - simplify.

flint/rmsynth.py
```diff
@@ -664,10 +701,9 @@ def write_rm_products(
         rmsynth_options (RMSynthOptions): Options controlling RM-synthesis
         rmclean_options (RMCleanOptions): Options controlling RM-CLEAN
         cube_products (list[FDFLabel]): Which FDF cube(s) to write ('dirty', 'clean', 'model')
-        moment_products (list[FDFLabel]): Which FDF(s) to compute Faraday moment maps from
+        moment_products (list[FDFLabel]): Which FDF(s) to compute Faraday moment maps from, twelve per entry
```
AlecThomson:
```suggestion
        moment_products (list[FDFLabel]): Which FDF(s) to compute Faraday moment maps from
```

AlecThomson (commented):
@claude again you add too much verbose text and unnecessary private global variables

flint/rmsynth.py
AlecThomson: No, as it turns out. Restoring `FaradayMoments` above left it with one use, so it is a local in `write_moment_maps_to_fits` now (7716a3a). The filter itself still earns its keep — it is what stops `mom0_debias` being written a second time under a second name — but not as a module-level table.

---

flint/rmsynth.py
AlecThomson: Fixed in e7c3e28. The tell was that `write_peak_maps_to_fits` had kept `FaradayPeaks` and is handed one rebuilt at the call site — only the moments path had been flattened. Both parameters take `FaradayMoments` again and the call site mirrors the peaks one.

One consequence: the tuple needs all twelve fields, so the debiased set now computes `mom0_debias` as well. It is still not written — the filter is purely a write filter now.

---

flint/rmsynth.py
AlecThomson: Cut to two lines in e7c3e28.

---

flint/rmsynth.py
AlecThomson: Done in e7c3e28. `PEAK_MAPS` too, and `DEBIASED_MOMENT_MAPS` with them for consistency — though that one has since become a local (7716a3a), so only the two remain.

---

flint/rmsynth.py
AlecThomson: Deleted in e7c3e28. The matching one over `PEAK_MAPS` went too (7716a3a) — leaving the twin behind made no sense.

---

flint/rmsynth.py
AlecThomson: Done in e7c3e28.

---

flint/rmsynth.py
AlecThomson: Applied in e7c3e28.

---

flint/options.py
AlecThomson: Applied in e7c3e28, as `'dirty', 'clean', 'model'`.

---

flint/options.py
AlecThomson: Applied in e7c3e28, as `'dirty', 'clean', 'model'`.

---

flint/options.py
AlecThomson: Applied in e7c3e28.

---

flint/options.py
AlecThomson: Applied in e7c3e28.

---

flint/options.py
AlecThomson: Applied in e7c3e28.

---

PR #423, opened 2026-09-10
Take up rm-lite 2026.9.3, and expose the zarr conversion

AlecThomson:
Companion to AlecThomson/rm-lite#81, #82, #83 and #84, which came out of a RACS field whose RM-synthesis stage died with `KilledWorker`.

rm-lite now takes the FDF's precision from the cubes it is handed rather than always using complex128. flint feeds it float32 cubes, so it gets complex64 and halves the memory of every Faraday-depth array for free.

## Pin

`rm-lite==2026.9.4`, which carries the precision and task-graph work (#81), the zarr IO (#82), the `convert_to_zarr` bool (#83) and `py.typed` (#84). `uv.lock` regenerated with `uv lock`.

That line raises rm-lite's floor to Python 3.11, because sharded zarr writes need zarr 3. flint already requires `>=3.11,<3.14`, so nothing changes here.

`py.typed` means mypy now type-checks flint's calls into rm-lite instead of skipping them as untyped. flint's mypy hook runs isolated without rm-lite installed, so CI is unaffected; the benefit shows up running mypy in a real environment.

## Expose the zarr conversion

`RMSynthOptions.convert_to_zarr` is a bool, passed straight to `rmsynth_3d_from_fits`. rm-lite converts each cube to a store beside it, named after it:

```
SB9999.beam00.q.cube.fits  ->  SB9999.beam00.q.cube.zarr
SB9999.beam00.u.cube.fits  ->  SB9999.beam00.u.cube.zarr
SB9999.beam00.i.cube.fits  ->  SB9999.beam00.i.cube.zarr
```

Worth setting for a cube wide enough that the chunking has to split the image width. A FITS cube is one contiguous array, so a block narrower than the image still costs a full-width read; a zarr chunk costs itself. On the cube rm-lite measured, one 0.2 MiB block cost 512 MiB resident from FITS against 11 MiB from a store. Off by default, since it copies the data.

Checked end to end through the real path rather than the mocked one: with the flag off no stores are written, with it on exactly those three appear, and `max |plain - converted| = 0.0`.

## Warn on a double-precision cube

If a float64 cube ever reaches the RM-synthesis stage, the FDF, RMSF and CLEAN cubes all become complex128 and every Faraday-depth array doubles. That would now be silent, so `check_cubes_are_single_precision` reads the BITPIX of each Stokes cube and warns. Warned rather than refused: it costs memory, it is not wrong, and rewriting a 16 GB cube is not something the pipeline should do on its own.

## Report the precision and the chunk shape

`_describe_rm_workload` now includes the FDF's dtype and the spatial chunk shape:

```
69 products over 17000x19533 pixels and 587 Faraday depths in complex64,
in 3400 chunks of 5x19533 pixels; per-pixel RMSF on
```

Both were invisible before. On the failing run the log said "in 17000 chunks" and nothing about how big one was, which is the number that actually explains a killed worker. `target_chunk_mb` was also being silently ignored from the FITS entry point (fixed in rm-lite), and this line is where that would have shown.

## Review follow-ups

- `convert_to_zarr` is a bool rather than a path. A strategy file is reused across fields, so a directory set there would have every field write over the last one's stores. Fixed properly in rm-lite #83 by naming stores after their cubes, which removes the failure mode for every caller rather than just this one.
- `check_cubes_are_single_precision`, `compute_rm_products` and `fdf_threshold_from_snr` dropped their underscores, since the tests import them. The last also says what it returns.
- `test_rmsynth_options_reach_rm_lite` now checks that every `RMSynthOptions` field reaches rm-lite, rather than the two it happened to assert. Its docstring already claimed the general property. `debias_moments` and `debias_filter_size` are named as the exceptions, since flint applies those itself after rm-lite returns.

`uv run pytest tests/test_rmsynth.py` passes (43 tests) against a clean `uv sync` of the new pin. `pre-commit run --all-files` clean.

https://claude.ai/code/session_01AN9y36hf41ZmfkLQPnd4AN

---
_Generated by [Claude Code](https://claude.ai/code/session_01AN9y36hf41ZmfkLQPnd4AN)_

### DIFF
--- a/flint/options.py
+++ b/flint/options.py
@@ -363,8 +363,8 @@ class RMSynthOptions(BaseOptions):
     """finufft OpenMP threads per dask chunk"""
     target_chunk_mb: float = 256
     """Target per-chunk memory footprint, in MB, when reading the Q/U cubes"""
-    convert_to_zarr: Path | None = None
-    """Directory to copy the cubes into as zarr stores before reading them, chunked as the FDF needs. Worth it for a cube wide enough that the chunking has to split the image width, since a FITS block narrower than the image still costs a full-width read. None reads the cubes where they are"""
+    convert_to_zarr: bool = False
+    """Convert the cubes to zarr beside the Stokes Q cube before reading them. Worth it for a wide image, where a FITS read is full-width whatever the chunk"""
     fit_order: int = 2
     """Stokes I fractional-polarisation fit order; negative iterates orders and picks the best by AIC"""
     fit_function: Literal["log", "linear"] = "log"
--- a/flint/options.py
+++ b/flint/options.py
@@ -364,7 +364,7 @@ class RMSynthOptions(BaseOptions):
     target_chunk_mb: float = 256
     """Target per-chunk memory footprint, in MB, when reading the Q/U cubes"""
     convert_to_zarr: bool = False
-    """Convert the cubes to zarr beside the Stokes Q cube before reading them. Worth it for a wide image, where a FITS read is full-width whatever the chunk"""
+    """Convert each cube to a zarr store beside it before reading, so cube.fits gives cube.zarr. Worth it for a wide image, where a FITS read is full-width whatever the chunk"""
     fit_order: int = 2
     """Stokes I fractional-polarisation fit order; negative iterates orders and picks the best by AIC"""
     fit_function: Literal["log", "linear"] = "log"
--- a/flint/options.py
+++ b/flint/options.py
@@ -363,6 +363,8 @@ class RMSynthOptions(BaseOptions):
     """finufft OpenMP threads per dask chunk"""
     target_chunk_mb: float = 256
     """Target per-chunk memory footprint, in MB, when reading the Q/U cubes"""
+    convert_to_zarr: Path | None = None
+    """Directory to copy the cubes into as zarr stores before reading them, chunked as the FDF needs. Worth it for a cube wide enough that the chunking has to split the image width, since a FITS block narrower than the image still costs a full-width read. None reads the cubes where they are"""
     fit_order: int = 2
     """Stokes I fractional-polarisation fit order; negative iterates orders and picks the best by AIC"""
     fit_function: Literal["log", "linear"] = "log"
--- a/flint/options.py
+++ b/flint/options.py
@@ -364,7 +364,7 @@ class RMSynthOptions(BaseOptions):
     target_chunk_mb: float = 256
     """Target per-chunk memory footprint, in MB, when reading the Q/U cubes"""
     convert_to_zarr: bool = False
-    """Convert each cube to a zarr store beside it before reading, so cube.fits gives cube.zarr. Worth it for a wide image, where a FITS read is full-width whatever the chunk"""
+    """Convert each FITS cube to a zarr store beside it before reading"""
     fit_order: int = 2
     """Stokes I fractional-polarisation fit order; negative iterates orders and picks the best by AIC"""
     fit_function: Literal["log", "linear"] = "log"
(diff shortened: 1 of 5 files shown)


flint/options.py
```diff
@@ -363,6 +363,8 @@ class RMSynthOptions(BaseOptions):
     """finufft OpenMP threads per dask chunk"""
     target_chunk_mb: float = 256
     """Target per-chunk memory footprint, in MB, when reading the Q/U cubes"""
+    convert_to_zarr: Path | None = None
+    """Directory to copy the cubes into as zarr stores before reading them, chunked as the FDF needs. Worth it for a cube wide enough that the chunking has to split the image width, since a FITS block narrower than the image still costs a full-width read. None reads the cubes where they are"""
```
AlecThomson: @claude - too verbose

tests/test_rmsynth.py
```diff
@@ -28,6 +28,7 @@
     PEAK_MAPS,
     FDFLabel,
     RMSynth3DResults,
+    _check_cubes_are_single_precision,
     _compute_rm_products,
     _snr_threshold,
```
AlecThomson: If we're importing them - if for tests, they probably shouldn't be prive

AlecThomson: `_snr_threshold` isn't self-descriptive

flint/options.py
```diff
@@ -363,6 +363,8 @@ class RMSynthOptions(BaseOptions):
     """finufft OpenMP threads per dask chunk"""
     target_chunk_mb: float = 256
     """Target per-chunk memory footprint, in MB, when reading the Q/U cubes"""
+    convert_to_zarr: Path | None = None
```
AlecThomson: @claude shouldn't this be a straight bool?

AlecThomson: Yes, and this was a real bug rather than a style problem. Thanks for catching it.

A strategy file is reused across fields, so a directory set in the options is one fixed path for every field it runs over. rm-lite writes fixed names into it (`q.zarr`, `u.zarr`, ...) with `overwrite=True`, so the second field would silently overwrite the first field's stores and then synthesise from them.

Now `convert_to_zarr: bool = False`, and `zarr_store_directory` names the store after the Stokes Q cube so each field gets its own.

---

flint/options.py
AlecThomson: Cut to one line:

```python
convert_to_zarr: bool = False
"""Convert the cubes to zarr beside the Stokes Q cube before reading them. Worth it for a wide image, where a FITS read is full-width whatever the chunk"""
```

---

tests/test_rmsynth.py
AlecThomson: Agreed. All three dropped their underscores, since the tests import them:

- `_check_cubes_are_single_precision` → `check_cubes_are_single_precision`
- `_compute_rm_products` → `compute_rm_products`
- `_snr_threshold` → `fdf_threshold_from_snr`

The first is mine from this PR, so that one was me overdoing the underscore again.

---

AlecThomson: Renamed to `fdf_threshold_from_snr`, which says what it returns: an absolute FDF amplitude cut derived from an SNR multiple of the theoretical noise.

---

flint/options.py
```diff
@@ -363,6 +363,8 @@ class RMSynthOptions(BaseOptions):
     """finufft OpenMP threads per dask chunk"""
     target_chunk_mb: float = 256
     """Target per-chunk memory footprint, in MB, when reading the Q/U cubes"""
+    convert_to_zarr: bool = False
+    """Convert each cube to a zarr store beside it before reading, so cube.fits gives cube.zarr. Worth it for a wide image, where a FITS read is full-width whatever the chunk"""
```
```suggestion
    """Convert each FITS cube to a zarr store beside it before reading"""
```

AlecThomson: Applied in `4dd4d82`.

The "where a FITS read is full-width whatever the chunk" reasoning is on rm-lite's own `convert_to_zarr` docstring, so it is a jump away rather than lost.

---

PR #428, opened 2026-09-14
Make the total-resolution cubes in the polarisation flow

AlecThomson:
## What

The polarisation flow writes **natural**-resolution cubes: one beam per channel, so resolution follows frequency. RM synthesis needs **total** resolution, a single beam across the whole band.

So far rm-synth has made that itself, in `_resolve_common_resolution_cubes`: it splits each finished cube back into 2D planes, convolves each, and restacks. In the racs-all flow that cube had just been built, so the split and restack was a round trip.

This convolves each co-added plane to the total beam while it is still in hand, and stacks it into a second cube. One extra 2D convolution per channel instead of 36, and no second linmos. rm-synth then finds its inputs already share a beam and passes them straight through.

`process_racs_all` turns this on itself whenever the rm-synth stage runs.

## Naming

Resolution is now part of the name rather than implicit, so both sets say what they are:

```
SB59058.RACS_1626-84.pol.natural.image.cube.fits
SB59058.RACS_1626-84.pol.total.image.cube.fits
```

`ResolutionModes` in `naming.py` grows `natural` and `total` alongside the existing `optimal`/`fixed`, and gains `raw`, which `get_beam_resolution_str` already accepted and `test_get_beam_resolution_str` already asserted. Every suffix now comes from that vocabulary rather than a bare string, which is what the two standing TODOs on `create_image_cube_name` were asking for.

This renames the existing pol cube products. That is deliberate and in scope on this WIP branch.

## rm-synth is unchanged

It still checks the cubes it is given and convolves them when they do not share a beam, so it stays runnable standalone. Inside racs-all it now no-ops. That hinges on one detail worth flagging for review:

`cubes_share_common_beam` returns `False` when any real beam exceeds the cutoff, because blanking is work only a convolution pass does. `convolve_plane_to_beam` stamps `BMAJ/BMIN/BPA = 0.0` on a plane beyond the cutoff, and `usable_beam_mask` treats a zero beam as unusable at any cutoff, so blanked channels drop out of both masks and the check passes. This is why the total path uses `convolve_plane_to_beam` rather than bare `convolve_images`, which does not stamp. `test_total_resolution_cubes_share_one_beam` pins it.

## Other notes

- `task_convolve_plane_to_beam` moves from `prefect/common/rmsynth.py` to `prefect/common/imaging.py`. Importing it the other way would have been circular, and that module is deliberately isolated so the optional `rm-lite` dependency stays out of flows that do not need it.
- `remove_original_images` has the natural image cube delete the planes the total convolution reads, so the cube task now waits on those convolutions. Previously that ordering was only incidental.
- BANE, when asked for, runs on both sets, so the noise handed to rm-synth describes the resolution the FDF is built at.
- Spice receives the total cubes and their BANE cubes, or they would sit untrimmed and uncompressed beside everything that was spiced.
- With the flag off the task graph is unchanged.

## Testing

New `tests/test_total_resolution.py` (5 tests) covers the total beam solve, the shared-beam guarantee above, and cutoff blanking. `tests/test_racs_all_pipeline.py` gains 5 wiring tests (30 pass). `tests/test_naming.py` gains the vocabulary and cube-name tests (43 pass). Ruff and ruff-format are clean.

The full suite was not run locally; leaving it to CI.

https://claude.ai/code/session_01AN9y36hf41ZmfkLQPnd4AN

---
_Generated by [Claude Code](https://claude.ai/code/session_01AN9y36hf41ZmfkLQPnd4AN)_


docs/quickstart/polarisation.md
```diff
@@ -54,12 +54,26 @@ Two options steer this:
...
+The natural cubes are the ones to archive; the total cubes are the ones to
+synthesise from. The `racs-all` flow sets this itself whenever the rm-synth stage
+runs, so that stage is handed cubes already at one beam and does no convolution
+of its own.
+
+Run standalone, the rm-synth stage still checks the cubes it is given and brings
+them to a total beam when they do not share one, writing its own `.total` cubes
+and leaving the inputs alone. See `RMSynthFieldOptions.beam_cutoff` to drop the
+coarsest channels from that solve rather than smoothing the whole band to reach
+them.
```
AlecThomson:
```suggestion

```

flint/prefect/common/imaging.py
```diff
@@ -1081,6 +1103,28 @@ def linmos_channel_groups_to_cubes(
             bkg_planes.append(task_getattr.submit(bane_maps, "bkg_image"))
             rms_planes.append(task_getattr.submit(bane_maps, "rms_image"))
 
+        if total_beam_shape is not None:
+            # Convolved here, while the plane is already in hand, rather than by
+            # splitting the finished cube back apart downstream
```
AlecThomson:
```suggestion
```

flint/naming.py
```diff
@@ -314,35 +314,29 @@ def create_imaging_name_prefix(
     return ".".join(names)
 
 
-ResolutionModes = Literal["optimal", "fixed"]
+ResolutionModes = Literal["optimal", "fixed", "raw", "natural", "total"]
```
AlecThomson: Should be StrEnum at this point

AlecThomson: Done in 6563903.

A member is now its own filename suffix, so most call sites dropped `get_beam_resolution_str` entirely and just use `ResolutionModes.TOTAL` / `.OPTIMAL` / `.FIXED`. The helper is left for the two things it still earns: appending a `marker`, and coercing a mode that is still a loose string (which is also what keeps raising `ValueError` on an unknown one).

`raw` is in the enum too — `get_beam_resolution_str` already accepted it and `test_get_beam_resolution_str` already asserted it, but the old `Literal` left it out.

---

PR #431, opened 2026-09-21
Read the CLEAN model's moments on the scale it is actually on

AlecThomson:
## The problem

`write_rm_products` handed all three FDF cubes to the same moment call. The dirty and clean cubes are per-RMSF amplitudes, but the CLEAN model is a list of component fluxes, so its moment maps were divided by an RMSF area they never carried and came out about 13x low for the RACS band.

Seen in CARTA: `model.mom0` peaked at 0.0009 Jy/beam against a `clean.peak_pi` of 0.012. `0.0009 x 13.04 = 0.012`, where 13.04 is the RMSF area in Faraday pixels.

## The change

Each label names its own units, from the label alone, so nothing is exposed to the user:

```python
def fdf_units_for(label: FDFLabel) -> FDFUnits:
    return "integrated" if label == "model" else "per_rmsf"
```

Applied to both the fused `faraday_maps` path and the debiased `_lazy_faraday_moments` path.

## BUNIT

`mom0` has the RMSF area divided out and `peak_pi` does not, so the two cannot both be written as plain `Jy/beam`. The peak maps now carry `Jy/beam/RMSF`, matching RM-Tools (*"quantities derived from the FDF amplitude ... have the same unit"*) and the GMIMS/STAPS Faraday moment maps, which quote M0 as `K rad m-2 RMSF-1`.

Note that flint's `mom0` goes one step further than GMIMS/STAPS and divides the RMSF area out, landing in flux units so mom0 equals the peak for a Faraday-simple source. That is the more useful convention here, but it does differ from the published moment maps, so it is worth being explicit in the header.

## Testing

- `test_model_moments_are_on_the_flux_scale` runs synth and CLEAN end to end and checks the written `model.mom0` lands on the same scale as `clean.peak_pi`. Verified it has teeth: reverting the routing makes it fail, restoring it makes it pass.
- `test_peak_maps_are_written_per_rmsf` asserts the peak maps' `BUNIT` from the written FITS headers. Previously only the peak map *filenames* were checked, never their units.

Both pass locally; ruff clean. The full suite is left to CI.

## This needs a rm-lite release first

Depends on AlecThomson/rm-lite#93, where `calc_faraday_moments` gains a required `fdf_units` parameter. The pin here is still `rm-lite==2026.9.10`, which has no `FDFUnits`, so **CI will fail on import until that is released and the pin bumped**. I tested against a local rm-lite checkout rather than bumping the pin to a version that does not exist yet.

## Not addressed here

The other half of the original report, `moment_threshold_snr=0` on the clean cube, is unchanged. With no cut, mom0 integrates rectified noise over the whole Faraday axis (a `sqrt(pi/2) * sigma * n_phi * dphi / rmsf_area` pedestal) and steps by up to ~2.4x at the CLEAN mask edge, since `sum|FDF|` is not invariant under CLEAN while `max|FDF|` is. That is usage rather than a bug, so it is left alone.

https://claude.ai/code/session_01QD2Vhes35vCE2dk4DeF6Aw

---
_Generated by [Claude Code](https://claude.ai/code/session_01QD2Vhes35vCE2dk4DeF6Aw)_


tests/test_rmsynth.py
```diff
@@ -1895,3 +1897,61 @@ def test_fused_faraday_maps_match_the_per_map_path(per_pixel_noise: bool) -> Non
...
+
+
+def test_model_moments_are_on_the_flux_scale(
+    tmp_path: Path, qu_cubes: tuple[Path, Path]
+) -> None:
+    """The CLEAN model is a list of fluxes, not per-RMSF amplitudes.
+
+    Reading it as per-RMSF divides mom0 by the RMSF area a second time, which
+    put the model maps a factor of ~13 (the RMSF area in Faraday pixels) below
+    the peak for the RACS band. Every pixel of the fixture holds the same
```
AlecThomson: That's strictly the RACS-all band, correct? Also, that value is not tested here, so why document it like this? Is quite confusing

AlecThomson: Right on both counts, and worse than that: the fixture is not RACS at all. `_make_qu_cubes` builds a synthetic 700-1300 MHz band, so quoting a RACS number in this test's docstring was doubly misleading.

The factor is not a property of a band either. It is `rmsf_area / delta_phi`, i.e. `1.0645 * fwhm_rmsf / delta_phi`, so it depends on the Faraday sampling you chose as much as on the frequency coverage. "~13" was what came out of a 744-1032 MHz simulation at `d_phi = 4`, which is neither this fixture nor a general statement.

Dropped in `eba93a1`. The docstring now says only what the test actually asserts:

```
The CLEAN model is a list of fluxes, not per-RMSF amplitudes.

Reading it as per-RMSF divides mom0 by the RMSF area a second time. Every
pixel of the fixture holds the same Faraday-simple source, so the cleaned
flux has to land on the peak's scale.
```

The factor itself is tested where it belongs, in rm-lite's `test_moments_per_rmsf_on_a_model_is_out_by_the_rmsf_area`, which asserts it symbolically against `fwhm * gaussian_integrand(1.0, fwhm=1.0) / delta_phi` rather than as a magic number.

---

AlecThomson:
## `Format` is red on `eba93a1`, and it is the base branch's failure

The rm-lite pin blocker is resolved: `rm-lite==2026.9.11` is published, `uv.lock` is updated, and the `FDFUnits` import no longer fails. On the current head the only failing hook is mypy, in a file this PR does not touch:

```
flint/configuration.py:576: error: Assignment to variable "error" outside except: block  [misc]
flint/configuration.py:577: error: Trying to read deleted variable "error"  [misc]
Found 2 errors in 1 file (checked 126 source files)
```

Everything else in the job passes, `ruff format`, `codespell` and `typos` included.

**`racs-all-pol` fails this on its own, with none of this PR's code.** Checked out `origin/racs-all-pol` (`7507ff3`) into a clean worktree and ran the same hook:

```
mypy....................................................................Failed
  flint/configuration.py:576: error: Assignment to variable "error" outside except: block  [misc]
  flint/configuration.py:577: error: Trying to read deleted variable "error"  [misc]
  Found 2 errors in 1 file (checked 126 source files)
```

Its own CI agrees. `racs-all-pol` is #390, and that PR has been red since `397ba53`:

| run | commit | result |
|---|---|---|
| 658 | `7507ff3` "Bump lite" (current base tip) | failure |
| 657 | `397ba53` "Catch validation errors" | failure |
| 656 | `446c651` "Bump rm-lite" | success |

So the last green commit on that branch is the one immediately before `397ba53`. This PR's diff against it is `flint/rmsynth.py` and `tests/test_rmsynth.py` only; `flint/configuration.py` is not in it. The mypy hook is pinned at `mirrors-mypy` `v2.0.0`, so this is not version drift either.

**Cause.** `except (TypeError, ValidationError) as error:` appears five times (lines 461, 501, 524, 542, 567). Python deletes the exception name when the `except` block exits, so mypy treats `error` as deleted; line 576 then rebinds and reads that same name:

```python
        for error in errors:
            logger.warning(error)
```

**Proposed patch**, no behaviour change:

```python
        for message in errors:
            logger.warning(message)
```

No fix for this exists anywhere yet, so there is nothing to port in. I have not made the change here, since it is an unrelated file and it belongs on `racs-all-pol` (#390), where it also unblocks everything else targeting that branch. Happy to add it to this branch instead if that is preferred.

Not re-running the job: the failure is deterministic, not a flake.

---

PR #434, opened 2026-09-22
Drain RM product futures on the loop that waits on them

AlecThomson:
Fixes the `RuntimeError: <asyncio.locks.Event object ...> is bound to a different event loop` that killed a RACS RM-synthesis run 1h49m in, 41 of 69 products through `compute_rm_products`.

## What went wrong

`compute_rm_products` called `as_completed(list(future_to_keys))` without a `loop`, and `as_completed` falls back to `default_client().loop`:

```python
self.loop = loop or default_client().loop
```

The client backing `task_write_rm_products` is built with `set_as_default=False`, which also means `Client.__enter__` skips `_current_client.set(self)` — so it is invisible to `default_client()`. On a worker the default is instead the worker's own client, which `Worker._get_client` builds with `loop=self.loop` and `set_as_default=True` and never closes. Meanwhile a sync `Client` created with no `loop=` (what `prefect_dask.get_dask_client` does) gets its own dedicated IO-loop thread.

So `as_completed` tracked the futures on the **worker's** loop while `.result()` waited on the **client's own**.

## Why it stayed hidden for 41 products

`asyncio.Event` binds its loop lazily, on the first `wait()` that actually has to block, and a set event short-circuits:

```python
async def wait(self):
    if self._value:
        return True      # never touches _get_loop()
```

`as_completed` only yields a future once it has finished, so every `.result()` in the drain loop hit that fast path and the mismatch never surfaced.

The first future dask had to genuinely re-wait raised instead. `Client._gather`, when it cannot fetch a key's data, clears the event and loops back into the wait:

```python
log("Couldn't gather %s keys, rescheduling %s", ...)
self.futures[key].reset()   # clears the Event
```

That next wait calls `_get_loop()`, sees the worker's loop, and raises — which matches the `[unset]` event in the traceback. A recoverable reschedule became a dead flow.

## The change

Pass the owning client's loop, so only one loop ever touches those events and the reschedule can do its job:

```python
for future in as_completed(list(future_to_keys), loop=scheduler.loop):
```

The deliberate `set_as_default=False` is left alone.

## Test

`test_futures_are_drained_on_the_owning_clients_loop` runs two live clients: a default one handed the cluster (sharing its loop, as the worker's client does) and the owning one connected by address with no loop (as `get_dask_client` does, getting its own loop thread). Two live clients are what makes the assertion mean anything — with only the owning one, `default_client()` resolves back to it and the wrong loop is the right loop by accident. The existing distributed tests use `Client(cluster)` with the default `set_as_default=True`, so both loops coincide and they cannot see this.

Verified the test fails on the unfixed code (and that the compute still *succeeds* there, which is the latent shape of the bug) and passes with the fix.

## Not fixed here

Something did make a key's data go missing around 06:46 — worker-side logs would say whether a worker died, and memory pressure is the obvious suspect for a 15300x19212x587 `complex64` run with per-pixel RMSF on. This change only stops that from being fatal.

## Checks run

`pytest tests/test_rmsynth.py` (56 passed) and `pre-commit run --all-files` (all hooks pass, mypy included). The full suite is left to CI.

Introduced by d03e56b (#406), the move from `dask.compute(scheduler=client)` to draining futures by hand.

https://claude.ai/code/session_01QD2Vhes35vCE2dk4DeF6Aw

---
_Generated by [Claude Code](https://claude.ai/code/session_01QD2Vhes35vCE2dk4DeF6Aw)_


flint/rmsynth.py
```diff
@@ -648,7 +648,17 @@ def compute_rm_products(
...
+        # `loop` has to be given. Left out, `as_completed` tracks the futures on
+        # `default_client().loop`, and the client backing a prefect task is
+        # deliberately not the default -- so that is the *worker's* loop, while
+        # `.result()` below waits on this client's own. An `asyncio.Event` binds
+        # to the first loop that blocks on it, so the mismatch stays hidden while
+        # every drained future is already finished: waiting on a set event never
+        # looks at the loop. The first one dask has to re-wait -- `Client._gather`
+        # rescheduling a key whose data went missing -- raises `is bound to a
+        # different event loop` instead, turning a recoverable retry into a dead
+        # flow.
```
AlecThomson: @claude - this is way to much commentary. This can be a single line

AlecThomson (commented):
@claude please reduce your comment volume by a lot

flint/rmsynth.py
AlecThomson: Eleven lines down to one, in `4c30898`:

```python
        # Without loop=, as_completed uses the worker's loop, not this client's.
        for future in as_completed(list(future_to_keys), loop=scheduler.loop):
```

That is the part the code cannot show. How the `asyncio.Event` binds lazily, and why 41 products got through before it bit, belong in the PR description, which already has them.

---

